PR #3, opened 2017-12-11
Fix link to style guide

bartdesmet:


anpaz-msft:
Hey Bart,
Thanks for the fix! Could you push a new (trivial) commit? The build was not triggered, it should be fixed now.

anpaz-msft:
Do you mind resubmitting the PR? Looks like the webhook for the build was not triggered correctly.
Thanks!

PR #21, opened 2017-12-16
Typo

kant:
Please be advised that unforunately at this point we are not ready 
to accept external contributions, but we are actively working on this 
so please stay tuned!


alan-geller:
Fixed in the recent release.

PR #35, opened 2018-03-03
Implements transformation-based synthesis in Q#

msoeken:
This is an Q# implementation of the unidirectional variant of the transformation based synthesis approach [Miller, Maslov, Dueck, *DAC* 2003].

There are two example operations that are using the automatic synthesis algorithms. The first example finds the circuit for the permutation and checks whether it correctly computes the permutation using exhaustive simulation of all input patterns. The second example implements an instance of the hidden-shift quantum algorithm, in which the bent function is the inner product applied to a two set of qubits, where the first set of qubits is permuted.

I tried to follow code conventions and documentation styles as I saw it in other samples.


msoeken:
I am closing this pull request since I open a request from a direct branch.

PR #47, opened 2018-04-19
Inline BindA and improve unit testing.

cgranade:
This PR inlines BindA to match the implementation of BindC and BindCA, as well as adds a unit test to guard against bugs that affect the body of BindA but not its adjoint body.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ matrix:
 script:
   - dotnet build QsharpLibraries.sln
   - dotnet test Samples/UnitTesting
-  - travis_wait dotnet test LibraryTests
+  - travis_wait 30 dotnet test LibraryTests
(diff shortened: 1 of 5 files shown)


cgranade:
I've updated the PR to extend the amount of time that Travis CI waits before ending the job, since we see occasional failures due to load. I've also updated the version of Electron used while I was at it, since that was generating some spurious warnings at times.

.travis.yml
```diff
@@ -23,4 +23,4 @@ matrix:
 script:
   - dotnet build QsharpLibraries.sln
   - dotnet test Samples/UnitTesting
-  - travis_wait dotnet test LibraryTests
+  - travis_wait 30 dotnet test LibraryTests
```
martinquantum: great catch! I am wondering if we have an increased time-out setting already in place for the VSTS tests?

PR #54, opened 2018-05-07
Fix quantum phase estimation

lemj3111:
Fix Quantum Phase Estimation algorithm (QFT to Inverse QFT) and add a simple unit test

### DIFF
--- a/LibraryTests/QuantumPhaseEstimationTests.qs
+++ b/LibraryTests/QuantumPhaseEstimationTests.qs
@@ -2,49 +2,48 @@ namespace Microsoft.Quantum.Tests {
 
     open Microsoft.Quantum.Primitive;
     open Microsoft.Quantum.Canon;
-	open Microsoft.Quantum.Extensions.Convert;
+    open Microsoft.Quantum.Extensions.Convert;
     open Microsoft.Quantum.Extensions.Testing;
-	open Microsoft.Quantum.Extensions.Math;
-
+    open Microsoft.Quantum.Extensions.Math;
 
     /// # Summary
     /// Assert that the QuantumPhaseEstimation operation for the T gate
-	/// return 0000 in the controlRegister when targetState is 0 and
-	/// return 0010 when the targetState is 1  
-    operation QuantumPhaseEstimation_Test () : ()
+    /// return 0000 in the controlRegister when targetState is 0 and
+    /// return 0010 when the targetState is 1  
+    operation QuantumPhaseEstimationTest () : ()
     {
         body
         {
-            let Oracle = DiscreteOracle(T_PhaseEstimation(_,_));
-            using ( QPhase = Qubit[5]){
-                let Phase = BigEndian(QPhase[0..3]);
-                let State = QPhase[4];
-                QuantumPhaseEstimation(Oracle,[State],Phase);
-                let ComplexOne = Complex(1.,0.);
-				let ComplexZero = Complex(0.,0.);
+            let oracle = DiscreteOracle(_TPhaseEstimation);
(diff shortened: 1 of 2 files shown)


LibraryTests/QuantumPhaseEstimationTests.qs
```diff
@@ -0,0 +1,54 @@
+namespace Microsoft.Quantum.Tests {
+
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+	open Microsoft.Quantum.Extensions.Convert;
```
cgranade: It looks like you have a mix of spaces and tabs here, would you be willing to normalize to four spaces?

LibraryTests/QuantumPhaseEstimationTests.qs
```diff
@@ -0,0 +1,54 @@
...
+    open Microsoft.Quantum.Extensions.Testing;
+	open Microsoft.Quantum.Extensions.Math;
+
+
+    /// # Summary
+    /// Assert that the QuantumPhaseEstimation operation for the T gate
+	/// return 0000 in the controlRegister when targetState is 0 and
+	/// return 0010 when the targetState is 1  
+    operation QuantumPhaseEstimation_Test () : ()
```
cgranade: Most of our unit tests have names that follow the CamelCase convention (capitalized, without underscores), following style guide suggestions for operation and function names more generally.

LibraryTests/QuantumPhaseEstimationTests.qs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        body
+        {
+            let Oracle = DiscreteOracle(T_PhaseEstimation(_,_));
```
cgranade: Our style guide suggests that variable names generally should follow the camelCase convention (no initial capitalization, no underscores).

cgranade: The `(_, _)` notation here for partial application is correct, but isn't actually needed; the bare operation name `T_PhaseEstimation` is a value whose type is `(Int, Qubit[]) => () : Adjoint, Controlled`. It might be good to remove the `(_, _)` notation here, since no partial application is actually being performed in this case.

LibraryTests/QuantumPhaseEstimationTests.qs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        body
+        {
+            let oracle = DiscreteOracle(_TPhaseEstimation);
+            using ( qPhase = Qubit[5]){
+                let phase = BigEndian(qPhase[0..3]);
+                let state = qPhase[4];
+                QuantumPhaseEstimation(oracle,[state],phase);
+                let complexOne = Complex(1.,0.);
+				let complexZero = Complex(0.,0.);
```
anpaz-msft: this is using tabs... could you make sure it's all blank spaces?

LibraryTests/QuantumPhaseEstimationTests.qs
```diff
@@ -0,0 +1,53 @@
...
+                for(idxPhase in 0..4){
+                    AssertQubitState((complexOne,complexZero),qPhase[idxPhase],0.000001);
+                }
+                X(state);
+                QuantumPhaseEstimation(oracle,[state],phase);
+                AssertQubitState((complexOne,complexZero),qPhase[0],0.000001);
+                AssertQubitState((complexOne,complexZero),qPhase[1],0.000001);
+			    AssertQubitState((complexZero,complexOne),qPhase[2],0.000001);
```
anpaz-msft: Same here. These are using tabs instead of spaces...

PR #55, opened 2018-05-09
Adds three OpenQASM samples; Q# to OpenQASM, OpenQASM to Q# and running Q#, by exporting OpenQASM on IBMQ

RolfHuisman:
Adding a quick and dirty poc to show as an example that Q# will enable people to run their programs on the IBM Q machines as well. Proving my point that its an universal language.


RolfHuisman:
ApiKey was already been revoked, so its not an issue that its part of the commit.

QuantumExperience/Operation.qs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Quantum.Quasm
+{
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+
+	//Puts a qubit in superposition using the Hadamar gate and measures it
+    operation Hadamar () : (Result)
```
martinquantum: Hadamard

QuantumExperience/Operation.qs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        body
+        {
+			mutable result = Zero;			// set a mutable field to put the result in
+			using (qubits = Qubit[1])		//Create a Qubit register
+            {
+				H(qubits[0]);				//Perform Hadamar on the Qubit,
```
martinquantum: same here: please replace with "Hadamard"

QuantumExperience/Driver.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Sample to show that one can substitue the operation factory 
+        /// to run on different types of machines.
+        /// </summary>
+        /// <param name="args"></param>
+        static void Main(string[] args)
+        {
+            //You need to replace this with your own key from the quantum experience
+            var apiKey = "4616efdc29c9d4d751b3cd23a2e7d677ef8a6ff22b693afe0352e4f42b63e3a1135dc368484400b53cf7f4c8b96dcabf7cfc08442f2a1623614734c2e11e25f9";
+            var factory = new IbmQx4(apiKey); //Using different Factory
+            Console.WriteLine("Hadamar on IBMQx4");
```
martinquantum: same here: Hadamard

QuantumExperience/Driver.cs
```diff
@@ -0,0 +1,32 @@
...
+            for (int i = 0; i < 1; i++)
+            {
+                var result = Hadamar.Run(factory).Result;
```
martinquantum: and here again: Hadamard.Run

QuantumExperience/Driver.cs
```diff
@@ -0,0 +1,32 @@
...
+                Console.WriteLine($"Result of hadamar is {result}");
```
martinquantum: Hadamard

QuantumExperience/QuasmDriver.cs
```diff
@@ -0,0 +1,171 @@
...
+                            return QVoid.Instance;
+                        }
+                        QuasmLog.AppendLine("h q[" + (uint)q1.Id + "];");
+                        return QVoid.Instance;
+                    };
+                }
+            }
+        }
+
+        public override void StartOperation(string operationName, OperationFunctor functor, object inputValue)
```
anpaz-msft: This is supposed to be triggered on every operation, not only user defined ones. The fact that calling X or H doesn't trigger it is a bug, which has been fixed internally.
I don't think you want to put this on every call. I would suggest putting it on the constructor of the simulator.

QuantumExperience/QuasmDriver.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            public QSimCCNOT(IOperationFactory m) : base(m)
+            {
+            }
+
+            public override Func<(Qubit, Qubit, Qubit), QVoid> Body
+            {
+                get
+                {
+                    throw new NotImplementedException();
```
anpaz-msft: I recommend removing all these NotImplemented operations.

QuantumExperience/QuasmDriver.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+            }
+
+            public override Func<Qubit, QVoid> Body
+            {
+                get
+                {
+                    return delegate (Qubit q1)
+                    {
+                        if (q1 == null)
```
anpaz-msft: doesn't this need to be logged in the quasm.log?

anpaz-msft:
Does it require Windows, or would it work on Mac/Linux? I don't see why it wouldn't.
The text in the README.md seems to imply that it only works on Windows.

anpaz-msft:
All code files need to have License added.

anpaz-msft:
It would be nice to have a README.md that explains installation and running instructions as well as a little explanation of what this sample does.

martinquantum:
This should work ok with any platform that supports python3. Otherwise, the IBM Experience is basically just a python project that can be imported by “from IBMQuantumExperience import IBMQuantumExperience” in the interface code QiskitInterface.py. And all you need to install the IBM SDK is “pip install qiskit”.

In Rolf’s code there is a bunch of stuff hardcoded (like the shell command “bash”) etc which is not very user friendly. Also, there are almost no comments.

Martin

QuantumExperience/QuasmDriver.cs
```diff
@@ -0,0 +1,171 @@
...
+                        {
+                            return QVoid.Instance;
+                        }
+                        QuasmLog.AppendLine("h q[" + (uint)q1.Id + "];");
```
anpaz-msft: Interpolated strings IMHO would make it a bit easier to read, i.e.: 
`QuasmLog.AppendLine($"h q[{q1.Id}];"); `

QuantumExperience/QuasmDriver.cs
RolfHuisman: Check, will be moving that.

QuantumExperience/QuasmDriver.cs
RolfHuisman: Fair, where place holders for later. Will remove them in this one.

QuantumExperience/QuasmDriver.cs
RolfHuisman: Not yet used in the example. But indeed; Should have been for YXZ.

QuantumExperience/QuasmDriver.cs
RolfHuisman: Would indeed make it more readable.

QuantumExperience/QuasmDriver.cs
RolfHuisman: Need to leave the measurement one in place though. Because of a reflection issue in the simulatorbase, if I leave it out the simulator thinks its not completely implemented.

qubeat:
I would suggest simply generate qasm file without try to send it to hardware. In such a case lot of problems disappear and even API key is not necessary (it may be ran on offline simulators, after all). Anyway it may not work for two-qubit gates without topology mapping information (taking into account connectivity and "direction" of C-NOT). Even for single qubit it is desirable to request availability and size of queue in device.

RolfHuisman:
I agree with you that just generating the file instead of sending it to the hardware would be easier to implement and probably look a lot cleaner. This pull request is a reduced and cleaned up version of what I normally use in my demo's of Q#. The number one question I got after doing the demo's of Liquid (Q# predecessor) is showing this on the real hardware. People understood that the QASM file coud run on the machine, but having it inline had a different perception. Having said that, maybe its a good idea to plit this into two samples. With a more fuller simulator that has a mapper and a more inline sample.

qubeat:
Yes, I see. I was aware about LiquID, but only after reading Q# description realized that now it may be used with IBM QX quite naturally. But, the current device is anyway may not be controlled by user in "real-time" and qasm is desirable intermediate step. After that it may be used with different offline and online simulators or remapped to sent that into appropriate and currently available device. So I considered a working example with full set of Q# quantum gates, implemented at least for generating qasm file, as a desirable possibility. In fact, just before reading of your request I doubted, if to ask about that in an alternative site suggested as feedback pipeline. Such example would be useful also to clarify structure of QuantumSimulator and SimulatorBase stuff – I may not find a description anywhere. It may be useful also to show users how to do such a thing with other devices and simulators. E.g., I myself used in my old simulator maximally simple file format to save and read circuits and could try use Q# to generate that without necessity to write qasm parser and converter.

RolfHuisman:
Its currently mostly a sample to inspire, but I think it can be more. The current usage of SimulatorBase currently works, but I think I need to drop to the IAbstractFactory to get many of the more advanced constructs correct. The only thing that I'm currently going back and forward with myself over is where to and how put the mapper.

qubeat:
Quite illustrative. Yet, I do not realize yet, how definitions in QSimR should look like - it would be necessary for most general one-qubit U3 gate on the chip.

RolfHuisman:
https://docs.microsoft.com/en-us/qsharp/api/prelude/microsoft.quantum.primitive.r?view=qsharp-preview#remarks
> When called with pauli = PauliI, this operation applies a global phase. This phase can be significant when used with the Controlled functor.
Appart from the radials vs exponents in the rotation matrixes, I'm still not clear on the PauliI if this is based on the https://docs.microsoft.com/en-us/qsharp/api/prelude/microsoft.quantum.primitive.r1?view=qsharp-preview rotation or something else. Simulations indicate its something else, but can't find the details of it. For the rest it would be either a switch method which would generate different rotations in Quam or a morge generelized one using U3(,,)

qubeat:
The controlled gates are coded via c-not and u3 in qelib.inc. Currently cu3 uses u3 with normalized unit determinant and cu1 is used for additional phase. However in comment to tutorial about the stuff there is note about plan to rework that and use single 4-parameters gate. If implementation of Controlled QSimR is required for construction such a class in SimulatorBase? Compositions of C-NOT and R are enough to implement any gate.

RolfHuisman:
Curious to see why the linux build on vsts failed. Don't have the rights to see why. Suspect its because of something on the build environment.

RolfHuisman:
@anpaz-msft could you please do a re-review ? The old one isn't current anymore.

Samples/OpenQasm/OpenQasmDriver.cs
```diff
@@ -0,0 +1,321 @@
...
+    /// <summary>
+    /// Quick and dirty Simulatorbase to write OpenQASM 2.0
+    /// Just enough to show that it would work
+    /// Please don't put this in production until its fully engineerd.
+    /// This code could eat your cat. So imagine what schrodinger has to say about that one.
+    /// </summary>
+    public abstract class OpenQasmDriver : SimulatorBase
+    {
+        private int operationDepth;
+        public override void StartOperation(string operationName, OperationFunctor functor, object inputValue)
```
anpaz-msft: The StartOperation and EndOperation are the only things I have a bit of concerned. On the upcoming version we will change the interface of these methods to receive different parameters, and the behavior will also change as today Primitive operations (like X or H) are not triggering these events, but now they will, and it feels to me this will break this code.
Could you instead use the constructor or the `RunOpenQasm` method instead?
If primitive operations triggering these methods is not a problem, we could leave it like this and I will just take care of updating the signature with the new version, or even better, we could just wait for the new version which should not take long.

RolfHuisman: Removed the dependency on StartOperation and EndOperation and put it in the constructor. I put it in the RunOpenQasm, I will need to either wrap it all the time when there are multiple measurements in there, or check if it already has been added. Next version needs to measure more qubits, so a single action can be re-used.

Samples/Qiskit/Qiskit.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <Content Include="requirements.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="README.md" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Canon" Version="0.2.1802.2202-preview" />
```
anpaz-msft: this one is unnecessary as you have a project reference below.

Samples/Qiskit/QiskitQasmRunner.cs
```diff
@@ -0,0 +1,88 @@
...
+    internal static class QiskitExecutor
+    {
+        /// <summary>
+        /// Use the linux subsystem, to run in a real linux and run the python code to execute the Quasm
+        /// </summary>
+        public static string RunQuasm(StringBuilder quasm, int qbits, string key, string backend, int shots)
+        {
+            try
+            {
+                var directory = "Qiskit";
```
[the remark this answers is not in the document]
anpaz-msft: More over, looks like `QiskitInterface.py` expects the file to be in the same folder, so if you try to do:
`dotnet run` 
from the `Samples\Qiskit` directory, it doesn't work. I think you probably want to just write/read everything from the current folder.

Samples/Qiskit/requirements.txt
```diff
@@ -0,0 +1,2 @@
+qiskit
+IBMQuantumExperience
```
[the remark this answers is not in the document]
RolfHuisman: Qiskit is a dependency of the IBMQuantumExperience, so it gets automaticly installed when installing the Quantum Experience.

RolfHuisman:
Split up in #59 and #58, so closing this one.

Issue #62, opened 2018-06-25
RobustPhaseEstimationTest seems to be flaky


RolfHuisman:
Currently running builds. Sometimes RobustPhaseEstimationTest failes unexpectedly.
Seen it in the travis run (https://api.travis-ci.org/v3/job/396045026/log.txt):
> [xUnit.net 00:00:15.9331775]     RobustPhaseEstimationTest [FAIL]
[xUnit.net 00:00:15.9351386]       Microsoft.Quantum.Simulation.Core.ExecutionFailException : Assertion failed. Expected :  -2.61799387799149 . Actual :  3.6646801022588 
[xUnit.net 00:00:15.9366258]       Stack Trace:
[xUnit.net 00:00:15.9383051]         /home/travis/build/Microsoft/Quantum/Microsoft.Quantum.Canon/Asserts/ClassicalAsserts.qs(23,0): at Microsoft.Quantum.Canon.AssertAlmostEqualTol.<get_Body>b__5_0(ValueTuple`3 _args)
[xUnit.net 00:00:15.9392174]            at Microsoft.Quantum.Simulation.Core.Operation`2.Apply(I args)
[xUnit.net 00:00:15.9396934]            at Microsoft.Quantum.Simulation.Core.Operation`2.Microsoft.Quantum.Simulation.Core.ICallable<I,O>.Apply(I args)
[xUnit.net 00:00:15.9400038]         /home/travis/build/Microsoft/Quantum/LibraryTests/QcvvTests.qs(70,0): at Microsoft.Quantum.Tests.RobustPhaseEstimationTest.<get_Body>b__13_0(QVoid _args)
[xUnit.net 00:00:15.9413737]         C:\build-agent\_work\2\s\src\simulation\Testing\TestOperation.cs(40,0): at Microsoft.Quantum.Simulation.XUnit.TestOperation.<get_TestOperationRunner>b__6_0(IOperationFactory sim)
[xUnit.net 00:00:15.9422095]         /home/travis/build/Microsoft/Quantum/LibraryTests/SimulatorTestTargets.cs(58,0): at Microsoft.Quantum.Tests.SimulatorTestTargets.QuantumSimulatorTarget(TestOperation opData)
[xUnit.net 00:00:15.9436640]       Output:
[xUnit.net 00:00:15.9571821]         The seed, operation pair is ("Microsoft.Quantum.Tests.RobustPhaseEstimationTest",16645195)
Failed   RobustPhaseEstimationTest
Error Message:
[...]

PR #63, opened 2018-06-25
Fix for #62: RobustPhaseEstimationTest is flaky

RolfHuisman:
Fix for #62
Space in the class name accidently made the test use a random value instead of using the fixed value required for the test.


martinquantum:
adding @anpaz-msft and @cgranade as reviewers

anpaz-msft:
Beat me to the punch. I noticed this yesterday.

PR #65, opened 2018-06-26
Change the stored seed to a seed based on a hash to improve build resiliance

RolfHuisman:
Thought that I will just put on the table. Changing the stored seed to a seed generator to make each test class to have its own seed, but still make it deterministic. Have a number of builds which seem to fail randomly, so this might help in making the builds more predictable. But we don't wan't all tests to use one seed, because that might hide a lot of bugs. So generate a seed per test class, but still use the same seed for the tests.


RolfHuisman:
Pulling this one back untill its more clear what keeps failing in the build.

RolfHuisman:
See if I can include #66 in this set.

LibraryTests/SimulatorTestTargets.cs
```diff
@@ -199,19 +199,21 @@ private void LogSimulatorSeed(TestOperation opData, QuantumSimulator sim)
         }
 
         /// <summary>
-        /// Returns a seed to use for the test run if it is stored and null otherwise
+        /// Returns a seed to use for the test run based on the class
         /// </summary>
-        private uint? RetriveStoredSeed(TestOperation opData)
+        private uint? RetriveGeneratedSeed(TestOperation opData)
```
anpaz-msft: typo: RetrieveGeneratedSeed

RolfHuisman:
Adding same kind seed generation to QuantumSimulatorTestTargets.cs. Don't want to make a dependency between these two or add a utility library yet (bit overkill for one method).

LibraryTests/SimulatorTestTargets.cs
RolfHuisman: Corrected. Running the test, after pass will commit and push.

PR #67, opened 2018-06-29
Update dependent libraries to latest stable

RolfHuisman:
Update of the test SDK and Json to latest stable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: csharp
-dotnet: 2.1.4
+dotnet: 2.1.301
 mono: none
 
 matrix:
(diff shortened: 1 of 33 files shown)


RolfHuisman:
Updated to match the latest visual studio patches.

RolfHuisman:
The lock file is a fun one: 
https://medium.com/@Quigley_Ja/everything-you-wanted-to-know-about-package-lock-json-b81911aa8ab8
So probably that means that there are even newer versions out. But the main thing is, that we don't have "vulnerable" npm versions.

RolfHuisman:
Seems that the upgrade, automatically set the default of the lock file:
https://docs.npmjs.com/files/package-locks
^ apparently means "upgrade to compatible version of this when able".

RolfHuisman:
Pinned the last versions, to the latest stable, like @cgranade did for the previous upgrade.

Issue #66, opened 2018-06-27
Unittest failures in build which don't seem related to code changes


RolfHuisman:
If I look through https://travis-ci.org/Microsoft/Quantum/pull_requests i see a lot of failed builds, based on test failures that don't seem to be related to the change:

E.g.
https://travis-ci.org/Microsoft/Quantum/builds/396937185, log: https://api.travis-ci.org/v3/job/396937187/log.txt
 RepeatUntilSuccessCircuitsTest, Seed: 525109616
 MeasureWithScratchTest, Seed: 1049631137
 BitFlipSampleWt1CorrectionTest, Seed: 3681483977
 EstimateFrequencyTest, Seed: 2566129342
 KDTest, Seed: 1627590700
 KnillDistillationNoErrorTest, Seed: 4166661619
 BayesianPEManualTest, Seed: 3270398307
 Pi4YInjectionTest, Seed: 223591028   
 SingleQubitProcessTomographyMeasurementTest, Seed: 42796896
https://travis-ci.org/Microsoft/Quantum/builds/396906712, log: https://api.travis-ci.org/v3/job/396906714/log.txt
 SuperdenseCodingTest, Seed: 2280605845
 RepeatUntilSuccessStatePreparationTest, Seed: 4294571069
 RepeatUntilSuccessCircuitsTest, Seed: 587399533
 MeasureWithScratchTest, Seed: 2431261675
 BitFlipSampleWt1CorrectionTest, Seed: 3472194629
 EstimateFrequencyTest, Seed: 4232379153
 KDTest, Seed: 2963729218
 Pi4YInjectionAdjointTest, Seed: 1062934322
 KnillDistillationNoErrorTest, Seed: 1606786275
 BayesianPEManualTest, Seed: 1717820587
 Pi4YInjectionTest, Seed: 4254770888

Based on trying to reproduce the seeds, they don't give issues.
Other builds also seem to have this behavior; 
https://travis-ci.org/Microsoft/Quantum/builds/396394544, log: https://api.travis-ci.org/v3/job/396394546/log.txt
https://travis-ci.org/Microsoft/Quantum/builds/396372255, log: https://api.travis-ci.org/v3/job/396372257/log.txt
https://travis-ci.org/Microsoft/Quantum/builds/396045024, log: https://api.travis-ci.org/v3/job/396045026/log.txt

So I'm hypothesising its caused by either; an order of test or something that's not disposed.
Needs more diagnosing.

[...]

PR #68, opened 2018-06-29
Don't merge. Debugging tests

RolfHuisman:
Validating hypothesis, that the race condition is in the init method.
Debugging #66


RolfHuisman:
Doesn't seem the issue.

Issue #48, opened 2018-04-21
Q# compilation fails if user name contains a blank space


aviralwal:
I am trying to run the  `TeleportationSample` in `Samples > 0.Introduction folder` of `QsharpLibraries` solution. After setting the project as `Startup project`, I perform the build operation `F5`. But the build fails with error as :- 
```
1>------ Build started: Project: Microsoft.Quantum.Canon, Configuration: Debug Any CPU ------
1>No executable found matching command "dotnet-C:\Users\deepa"
[...]

PR #70, opened 2018-06-30
Updating to latest qdk version.

anpaz-msft:
This one does actually fix the user's name with blank space problem (#48). Sigh.


tcNickolas:
Please resolve conflicts so that CI can pick up the right version of the code to test.

PR #71, opened 2018-06-30
Update to latest IBMQuantumExperience interface

RolfHuisman:
Upgrade to the latest interface (which doesn't always give an expected queue size).

### DIFF
--- a/Samples/Qiskit/Driver.cs
+++ b/Samples/Qiskit/Driver.cs
@@ -24,12 +24,15 @@ namespace Microsoft.Quantum.Samples.Qiskit
                 Console.Error.WriteLine("Did you put an api key in Driver.cs ? Without that, it will not work.");
                 return;
             }
-            var factory = new IbmQx4(apiKey); //Using different Factory
-            Console.WriteLine("Hadamard on IBMQx4");
-            for (int i = 0; i < 1; i++)
+            else
             {
-                var result = MeasurementOneQubit.Run(factory).Result;
-                Console.WriteLine($"Result of Hadamard is {result}");
+                var factory = new IbmQx4(apiKey); //Using different Factory
+                Console.WriteLine("Hadamard on IBMQx4");
+                for (int i = 0; i < 1; i++)
+                {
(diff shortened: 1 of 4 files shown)


RolfHuisman:
Simplified the code to just wait for the timeout. Much simpler and more reliable. Sometimes the environment is fast enough that this just works.

Samples/Qiskit/Driver.cs
```diff
@@ -22,14 +22,16 @@ static void Main(string[] args)
...
-            var factory = new IbmQx4(apiKey); //Using different Factory
-            Console.WriteLine("Hadamard on IBMQx4");
-            for (int i = 0; i < 1; i++)
+            else
             {
-                var result = MeasurementOneQubit.Run(factory).Result;
-                Console.WriteLine($"Result of Hadamard is {result}");
+                var factory = new IbmQx4(apiKey); //Using different Factory
+                Console.WriteLine("Hadamard on IBMQx4");
+                for (int i = 0; i < 1; i++)
```
anpaz-msft: ?

Samples/Qiskit/Qiskit.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Content Include="QiskitInterface.py">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
anpaz-msft: ?

RolfHuisman: Accidental inclusion. Will correct.

Samples/Qiskit/Driver.cs
RolfHuisman: Moved the if statement, so it halts after the message. Otherwise one wouldn't see it.
The diff, makes it look wierd.

kok11:
Ok

RolfHuisman:
@kok11 thanks.

anpaz-msft:
Somehow I missed this had been updated. Sorry for the delay.

PR #72, opened 2018-06-30
Workaround for travis environment for suspected OpenMP concurrency bug on ubuntu 14.04

RolfHuisman:
Testing a number of hypothesis of #66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: csharp
-dotnet: 2.1.4
+dotnet: 2.1.301
 mono: none
 
 matrix:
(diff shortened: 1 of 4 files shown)


RolfHuisman:
Seems #66 is a concurrency bug in the dependencies used on ubuntu 14.04.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: csharp
-dotnet: 2.1.4
+dotnet: 2.1.301
```
RolfHuisman: Version 2.1.300 had an introduction of node re-use which caused a vstest issue that was fixed in 2.1.301. Using this version fixes some other dotnet test behaviors.

.travis.yml
```diff
@@ -13,6 +13,8 @@ matrix:
     # the APT source provided by Travis CI for more recent versions
     # of GCC backported to Ubuntu 14.04 Trusty.
     - os: linux
+      script:
+         - export OMP_NUM_THREADS=1
```
RolfHuisman: Limit to 1 thread on openmp. Seems to work around the current issue

PR #84, opened 2018-09-27
Add CHSH Game to Samples

ahelwer:
The CHSH Game is a famous nonlocal xor game which proves that no theory of local hidden variables can ever reproduce all of the predictions of quantum mechanics. This sample is a simple Monte Carlo simulation demonstrating the quantum strategy outperforming the best possible classical strategy for the CHSH Game over thousands of trials. It has documentation which includes a detailed description of the CHSH Game's workings as well as an account of its historical importance.

For educational value specific to the Q# language itself, this sample demonstrates measurement in nonstandard bases: the computational Z basis rotated +/- pi/8 radians around the unit circle.


Samples/CHSHGame/Game.qs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    operation MeasureBobQbit(bit : Bool, qubit : Qubit) : (Result)
+    {
+        body
+        {
+            if (bit)
+            {
+                // Measure in -pi/8 basis if bit is 1
+                Ry(0.78539816339, qubit);
```
tcNickolas: Nit: I would suggest to use constant `Microsoft.Quantum.Extensions.Math.Pi` here instead of a number. Not that I expect the value of Pi to change in the foreseeable  future... :-)

Samples/CHSHGame/Game.qs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                // Measure in pi/8 basis if bit is 0
+                Ry(-0.78539816339, qubit);
+                return M(qubit);
+            }
+        }
+    }
+
+    operation PlayQuantumStrategy(
+		aliceBit : Bool,
```
tcNickolas: Please use spaces instead of tabs

PR #87, opened 2018-10-01
Update of qiskit solution

RolfHuisman:
Some fixes for session on https://techorama.nl/
Latest qiskit version, removes need for linux subsystem on windows.
Support for, IbmQx16Melbourne, which I use because the ibmqx4 is having api issues (suspecting its going into maintenance soon).


RolfHuisman:
Lot of changes that don't belong in this pull request,.

PR #97, opened 2018-10-24
Sample using ReVerC to compile Q# code within a Q# project

meamy:
I've added a sample which uses [ReVerC](https://github.com/msr-quarc/ReVerC) to compile F# code to Q# circuits and use them within the same project. The example given compiles a classical F# adder to a Q# circuit and uses it to compute 1 + 1.

The main things to watch out for are the included DLLs, since fstarlib.dll is non-trivial to obtain from source now, and the hard-coded path to fsi.exe in the pre-build events, as fsi is not automatically added to any environment variables.


Samples/ReVerC/Driver.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Quantum.ReVerC
+{
+    class Driver
+    {
+        static void Main(string[] args)
+        {
+            using (var sim = new QuantumSimulator())
+            {
+                QArray<Result> a = new QArray<Result> { Result.One, Result.Zero };
```
cgranade: Would you be willing to add some comments here explaining how these two arrays represent the input into the `add` operation? Also, a quick C# hint, but you can use `var a` here to avoid repeating the type name `QArray<Result>` in the declaration and constructor call.

Samples/ReVerC/Operation.qs
```diff
@@ -0,0 +1,44 @@
+﻿namespace Quantum.ReVerC
+{
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+
+  /// Prepares a quantum register in the given basis state
+  operation prepare(initial: Result[], register: Qubit[]) : ()
```
cgranade: Q# naming conventions follow general .NET conventions, in that operation and function names follow `CamelCase` capitalization. Would you be willing to change the names of `prepare` and `add` to `Prepare` and `Add` accordingly?

Samples/ReVerC/Operation.qs
```diff
@@ -0,0 +1,44 @@
+﻿namespace Quantum.ReVerC
```
cgranade: The existing samples have been placed under the `Microsoft.Quantum.Samples` namespace, would you be willing to rename the namespace to `Microsoft.Quantum.Samples.ReVerC`? Thanks!

Samples/ReVerC/Operation.qs
```diff
@@ -0,0 +1,44 @@
...
+{
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+
+  /// Prepares a quantum register in the given basis state
```
cgranade: Would you be willing to normalize indents to four spaces?

Samples/ReVerC/Operation.qs
```diff
@@ -0,0 +1,44 @@
...
+  operation prepare(initial: Result[], register: Qubit[]) : ()
+  {
+    body
+    {
+      for (i in 0..Length(initial)-1)
```
cgranade: Generally, we try to avoid single-letter variable names. I'd suggest using `idxQubit` here to emphasize what is being indexed.

Samples/ReVerC/AdderGen.fsx
```diff
@@ -0,0 +1,31 @@
...
+        result.[i]  <-  a.[i] <> b.[i] <> carry
+      result           
+    @>
+
+/// Compiles a 2-bit adder (compile (carryRippleAdder 2)) and outputs the
+/// circuit in Q# format with the operation name "adder" to the file "adder.qs".
+///
+/// By convention, a function with type "a -> ... -> b -> c" is compiled to an
+/// operation with of type "a -> ... -> b -> c -> ()".
+File.WriteAllText("adder.qs", printQSharp "adder" <| compile (carryRippleAdder 2) false Eager)
```
cgranade: Is this where the name for the Q# operation `adder` is defined? If so, I'd suggest changing it to be CamelCase as well.

Samples/ReVerC/Operation.qs
```diff
@@ -0,0 +1,44 @@
...
+      mutable res = new Result[2];
+      using (qubits = Qubit[6])
+      {
+        let a = qubits[0..1];
+        let b = qubits[2..3];
+        let c = qubits[4..5];
+        prepare(ainit, a);
+        prepare(binit, b);
+
+        adder(a, b, c);
```
cgranade: Would you be willing to add some comments explaining that `adder` here is drawn from the ReVerC compilation? Thanks!

Samples/ReVerC/AdderGen.fsx
cgranade: Where is the `printQSharp` function defined?

Samples/ReVerC/AdderGen.fsx
```diff
@@ -0,0 +1,31 @@
...
+        carry <- compute_carry a.[i-1] b.[i-1] carry
+        result.[i]  <-  a.[i] <> b.[i] <> carry
+      result           
+    @>
+
+/// Compiles a 2-bit adder (compile (carryRippleAdder 2)) and outputs the
+/// circuit in Q# format with the operation name "adder" to the file "adder.qs".
+///
+/// By convention, a function with type "a -> ... -> b -> c" is compiled to an
+/// operation with of type "a -> ... -> b -> c -> ()".
```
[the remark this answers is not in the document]
meamy: Right, I forgot about that, ReVerC does actually uncurry the types. I've fixed the documentation.

Samples/ReVerC/Operation.qs
meamy: ReVerC compiles to the namespace `Quantum.ReVerC`, so barring a change on that end (which is certainly possible), Adder.qs will use the namespace `Quantum.ReVerC`.

A simple solution would be for both Driver and Operation to use the namespace `Microsoft.Quantum.Samples.ReVerC` and open `Quantum.ReVerC`. The C# compiler seems to be doing that fine, but the Q# compiler seems to think `Quantum.ReVerC` is a sub-namespace of `Microsoft.Quantum` and isn't able to import it.

As it is I've left Driver using the `Microsoft.Quantum.Samples.ReVerC` namespace, and Operation and Adder using `Quantum.ReVerC`. Let me know if that's not ideal and I can push an update to ReVerC!

Samples/ReVerC/AdderGen.fsx
meamy: `printQSharp` is a library function of ReVerC. I've switched to using qualified names to be clear about what ReVerC functions are used -- especially helpful since ReVerC is not well documented.

guenp:
@meamy Thank you for this contribution! It's been a while so will close the PR, feel free to reopen if you'd like to pick it back up again.

PR #101, opened 2018-10-28
clarifications to the English in explanations

Tranquilled:


Interoperability/python/qsharp/tomography.py
```diff
@@ -39,7 +39,7 @@ def single_qubit_process_tomography(
         prep = qsharp.Pauli.random()
         meas = qsharp.Pauli.random()
 
-        # Convert into a QuTiP object by using the standard transformation
+        # Convert tensor into QuTiP object by using the standard transformation
```
cgranade: The word "tensor" in `qt.tensor` isn't referring to a tensor, but to the tensor product. The conversion is thus not from a tensor to a Qobj, but from a pair of Q# `Pauli` values to a Qobj representing preparing and measuring the respective Pauli matrices.

cgranade:
```suggestion
        # Convert the preparation and measurement operators into a QuTiP object
        # by using the tensor product to represent the standard transformation
```

cgranade:
@Tranquilled: Thanks again for your contribution! I'm closing this PR for now, since it's been a little while, but please feel free to reopen it.

PR #103, opened 2018-10-29
Added another shortcut key

Gitcat711:


anpaz-msft:
`Ctrl + D` actually doesn't work on windows, and we're revamping the entire sample as the Python interop now works completely different (see https://docs.microsoft.com/en-us/quantum/install-guide/python). For now, I'm just closing this.

PR #112, opened 2018-11-09
Math Function - isPrime

jean002:
Here's a math function that checks if an integer different from -1, 1, 0 is prime.


martinquantum:
closing this PR.

PR #117, opened 2018-11-18
Guessing

jean002:
Here's an algorithm that can tell which element was picked in the set {0,1}, where each element has equal probability to be picked.


martinquantum:
closing this PR.

PR #147, opened 2019-03-01
Added Dockerfile with minimal install on Linux.

RudiP4t1qbit:


cgranade:
Thanks for adding this, @RudiP4t1qbit! I'm having some trouble getting it to work locally when testing, and wanted to check if I was running it wrong before bringing in the PR. In particular, using the command listed in `README.md`, the kernel dies with the error 'No executable found matching command "dotnet-iqsharp".' Thanks for the contribution and for your help!

RudiP4t1qbit:
@cgranade I wasn't able to reproduce this on my or Valentin's computer. That executable is installed in `/root/.dotnet/tools`, so the only time I've seen this break is when I accidentally mounted a directory to `/root` instead of a subdirectory the executable was gone.

If you reproduce that, can you post the output of `echo $PATH`? It should contain `/root/.dotnet/tools`. The directory should also contain an executable.

cgranade:
I'm a bit confused, as when I use `docker run -it rudip4t1qbit bash`, I can run `dotnet iqsharp --version` fine, and can confirm that `/root/.dotnet/tools` is on `$PATH` (see below), but I still get the same failure when running the notebook itself.

```bash
cgranade@MyDockerVM:~/Source/Repos/Quantum/DockerImages$ docker run -it rudip4t1qbit bash
[root@d3128c630ba3 ~]# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/root/.dotnet/tools
[root@d3128c630ba3 ~]# dotnet --version
2.2.104
[root@d3128c630ba3 ~]# dotnet iqsharp --version
Language kernel: 0.5.1902.2802
Jupyter core: 1.0.0.0
[root@d3128c630ba3 ~]#
```

cgranade:
One approach to help fix the issue I've been running into might be to update the IQ# kernel to use the latest Microsoft.Jupyter.Core package (v1.1), as we've added a new option with Microsoft/jupyter-core#14 to allow for explicit paths when adding kernelspecs. We added that installation option to help with issues we saw in CI, where it can be hard to get `$PATH` and .NET Global Tools to play along in a hosted environment.

(Adding @anpaz-msft to the thread for awareness.)

RudiP4t1qbit:
Hey @cgranade, I haven't been able to reproduce this issue. Which version of Docker are you using?

cgranade:
@RudiP4t1qbit: Thank you so much for posting this Dockerfile! Based on that, @anpaz-msft was able to get the binder configuration in #158 working, so we really appreciate the help there. I'm going to go on and close this for now, as your work on the Dockerfile has been included as a part of finishing up #158. Thank you again! 💕

PR #160, opened 2019-03-28
Updated tomography sample.

cgranade:
This PR uses the functionality in Microsoft/QuantumLibraries#47 to update the existing tomography sample and use the new IQ#-based Python interoperability library. This PR is in draft until 47 is merged in, at which point I'll rerun the notebook here and finish the README.


Samples/src/PythonInterop/tomography-sample.ipynb
```diff
@@ -29,482 +29,238 @@
...
-   ]
-  },
-  {
-   "cell_type": "markdown",
+   "execution_count": 10,
    "metadata": {},
+   "outputs": [],
    "source": [
-    "## Tomography ##"
+    "experiment = qsharp.compile(\"\"\"\n",
```
anpaz-msft: Should we move the definition of the operation into the `Operations.qs`? I don't see the need to compile it on the fly.

Samples/src/PythonInterop/environment.yml
```diff
@@ -9,20 +9,18 @@
 name: qsharp-samples
 channels:
     - conda-forge
-    - pythonnet
-    - https://solidrepo.blob.core.windows.net/release/0.3.1811.203/Libraries/Interoperability/python/bin/release/package
 dependencies:
     - python>=3.6
```
anpaz-msft: I just got a warning saying we should explicitly include `pip` since we're installing pip packages.

Samples/src/PythonInterop/tomography-sample.ipynb
cgranade: My thought was mainly to demonstrate on-the-fly compiling, but I agree it's not technically necessary in this case.

Samples/src/PythonInterop/environment.yml
cgranade: Ah, right, good point. I'll fix that, then.

PR #175, opened 2019-04-19
Oracle emulation sample

jgukelberger:
This adds a sample showing how oracles defined in terms of classical functions can be emulated by the QDK's simulator. Compared to a definition in terms of primitive gates, this allows for massive savings of auxiliary qubits and simulation time. Also, creation of an oracle from a classical permutation function becomes a one-liner, allowing for rapid prototyping.


msoeken:
This looks interesting! Maybe one can combine it with the `OracleSynthesis` or `ReversibleLogicSynthesis` samples to support hybrid simulation and synthesis, i.e., it's fast for simulation but can also be used for resource estimation.

jgukelberger:
> 
> 
> This looks interesting! Maybe one can combine it with the `OracleSynthesis` or `ReversibleLogicSynthesis` samples to support hybrid simulation and synthesis, i.e., it's fast for simulation but can also be used for resource estimation.

Looking at `ReversibleLogicSynthesis` that should be straight forward, since the input in both cases is a permutation table. Do you have a nice use case in mind where one could demonstrate costing and testing of a larger permutation oracle?

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,162 @@
...
+    open Microsoft.Quantum.Extensions.Emulation;
+
+
+    //////////////////////////////////////////////////////////////////////////
+    // Defining and using simple emulated oracles ////////////////////////////
+    //////////////////////////////////////////////////////////////////////////
+
+    // Declare an `intrinsic` oracle that is implemented in C#. See the
+    // `PermutationOracle.Register` call in the driver for its implementation.
+    operation HalfAnswer(x: Qubit[], y: Qubit[]) : Unit {
```
cgranade: This as well should probably have a Q# implementation rather than only providing a definition when run with the emulation feature.

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,162 @@
...
+    // ## oracle
+    // A quantum operation that implements an oracle
+    //      $O: \ket{x}\ket{y} \rightarrow \ket{x}\ket{f(x, y)}$.
+    operation RunConstantOracles (oracle: ((Qubit[], Qubit[]) => Unit)) : Unit {
+        Message("Querying the oracles...");
+
+        // Prepare a one-qubit register `x` and an eight-qubit register `y`.
+        // Since all the oracles here ignore x, its length and state do not
+        // matter.
+        using (qubits = Qubit[9]) {
```
cgranade: Generally, it's easier to read and maintain to allocate qubit variables directly using tuple deconstruction:
```suggestion
        using ((x, y) = (Qubit(), Qubit[8])) {
```
I'd also suggest more descriptive names for each register as well.

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,162 @@
...
+            oracle([x], y);
+            MeasureAndDisplay("The answer is ", y);
+
+            // Apply an oracle that was declared as `intrinsic` above and
+            // implemented in the C# driver.
+            HalfAnswer([x], y);
+            MeasureAndDisplay("Half the answer is ", y);
+
+            // Apply an oracle defined in terms of a Q# permutation function.
+            EmulateOracle(DoubleAnswerFunc, [x], y);
```
cgranade: I'd suggest asking @alan-geller for more advice about this line, but my initial impression is that this breaks the operational interpretation of `EmulateOracle` as something that could in principle be written as a set of instructions to a quantum device, as it takes a Q# function and implicitly applies that to the _state_ of its qubit arguments through the use of the C# definition above. If I understand correctly, an operationally consistent interpretation of the Q# program would still be possible, though, if the oracle function was defined at the C# level rather than the Q# level, such that from the perspective of the Q# program, the oracle would be a black box.

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+
+    //////////////////////////////////////////////////////////////////////////
+    // Emulated arithmetic operations ////////////////////////////////////////
+    //////////////////////////////////////////////////////////////////////////
+
+    // Prepare two `LittleEndian` registers in a computational basis state.
+    operation PrepareSummands(qubits: Qubit[], numbers: Int[]) : (LittleEndian, LittleEndian) {
```
cgranade: In general, the [Q# style guide](https://docs.microsoft.com/quantum/contributing/style-guide) suggests that qubit arguments should follow classical arguments. I'd also suggest that if the `qubits` input is always interpreted as two summands, that `(Qubit[], Qubit[])` may be a better type to represent that.

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,162 @@
...
+    // one-line function above. This defines already a permutation function on
+    // the computational basis states of two 8-qubit registers:
+    //      $\ket{m}\ket{n} \rightarrow \ket{m}\ket{m + n \mod 8}$.
+    // We can hence directly turn this function into an emulated oracle.
+    operation RunAddOracle() : Unit {
+        Message("Running emulated addition...");
+
+        // Turn the permutation function into an oracle operation acting on two
+        // quantum registers.
+        let adder = EmulateOracle(ModAdd8, _, _);
```
cgranade: Similar comment with respect to operationalism here.

msoeken:
> > This looks interesting! Maybe one can combine it with the `OracleSynthesis` or `ReversibleLogicSynthesis` samples to support hybrid simulation and synthesis, i.e., it's fast for simulation but can also be used for resource estimation.
> 
> Looking at `ReversibleLogicSynthesis` that should be straight forward, since the input in both cases is a permutation table. Do you have a nice use case in mind where one could demonstrate costing and testing of a larger permutation oracle?

In `ReversibleLogicSynthesis` I have been using an extended hidden shift example that also applies a permutation (once the original permutation and once its adjoint). This example can be easily scaled to different number of qubits. I can have a look once this is merged.

Samples/src/OracleEmulation/Emulator.qs
```diff
@@ -0,0 +1,28 @@
...
+    /// ## oracle
+    /// A function that defines the action of the oracle on the computational
+    /// basis states of the two registers x, y. The mapping
+    ///		$(x, y) \rightarrow (x, z=f(x, y))$
+    /// must be a bijective mapping on the basis states.
+    /// ## xbits
+    /// Input register x.
+    /// ## ybits
+    /// Output register y.
+    operation EmulateOracle(oracle : ((Int, Int) -> Int), xbits : Qubit[], ybits : Qubit[]) : Unit
```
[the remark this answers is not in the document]
jgukelberger: Thanks for the detailed review, Chris! I'll address the more stylistic comments later and will first reply to the general design points.

Regarding the point raised here, I agree `EmulateOracle` is essentially an intrinsic operation specific to one target machine. And, while there may be strong arguments against machine-specific constructs in Q# _on the language level_, I'm not sure we should rule out optimized Q# operations based on machine-specific primitives/intrinsics. In this case it is a simulator backend, but in the future it might be hardware capable of, say, a special multi-qubit operation that I'd like to use directly in my algorithm. Pretending that any piece of Q# code may be run on any target machine would ignore the fact that we'll have to think about the strengths and limitations of our target machine for quite some time.

NB: In principle, any call to `EmulateOracle` could be automatically mapped to a generic target machine by running a synthesis algorithm (as @msoeken is alluding to), but that's not necessarily what you want to do in practice.

Samples/src/OracleEmulation/Operations.qs
jgukelberger: Strictly requiring a Q# implementation with equivalent effects as the emulated version would 

1. preclude the rapid-prototyping benefits of oracle emulation and
2. introduce a non-checkable constraint (in general you'd have to be able to run the circuit to decide whether it's equivalent to a given permutation).

If we were to only require some Q# implementation (not necessarily equivalent to the emulated version), we'd likely end up with nonequivalent implementations, where the Q# version just fails, is a trivial or buggy operation because the user only cared about the emulated version. That does not appear desirable to me.

That being said, one indeed gets the greatest benefits from emulation if one can combine it with a circuit implementation for resource estimates. We will need more samples for that mode of operation. I am preparing one such sample in the context of arithmetic operations. @msoeken suggested another nice combination with his synthesis samples. This PR is meant to demonstrate just the basic ideas for defining and using an emulated oracle. Other are to be built on top.

Samples/src/OracleEmulation/Operations.qs
jgukelberger: I don't see a reason why this could not be made to run on a quantum device in principle: `EmulateOracle` is an operation, which could run, say, a reversible synthesis algorithm on the permutation function and then apply the corresponding circuit to its qubit arguments. For all the caller knows, it may act on the state of its qubit arguments by applying a series of gates - just as any other operation does.

Samples/src/OracleEmulation/Emulator.qs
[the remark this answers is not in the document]
jgukelberger: Tagging @alan-geller @bettinaheim @thomashaener @vadym-kl @matthiastroyer, who might have thoughts on machine-specific intrinsics. I realize this might establish some precedence, even if this is only in a sample for the moment.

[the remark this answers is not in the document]
jgukelberger: Discussing with @vadym-kl, it might be nice to have something like a `PermutationOracle` in `Canon` that implements the current interface of `EmulateOracle` via a synthesis algorithm. That would bring similar rapid prototyping benefits to generic target machines (with cost exponential in the register sizes of course); and it would most clearly demonstrate that a call to `EmulateOracle` could be implemented on any target machine.

Of course, more advanced synthesis algorithms again raise similar questions concerning choice of an optimal backend implementation depending on target machine capabilities.

Samples/src/OracleEmulation/Emulator.qs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+namespace Microsoft.Quantum.Extensions.Emulation
+{
+    /// # Summary
```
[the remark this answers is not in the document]
jgukelberger: Done!

Samples/src/OracleEmulation/Emulator.qs
```diff
@@ -0,0 +1,28 @@
...
+    /// Emulate the effect of a classical oracle by permuting the basis states of the simulator's
+    /// wavefunction such that
+    ///     $\ket{x}\ket{y}\ket{w} \rightarrow \ket{x}\ket{f(x, y)}\ket{w}$,
```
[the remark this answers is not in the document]
jgukelberger: Done!

Samples/src/OracleEmulation/Emulator.qs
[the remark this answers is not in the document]
jgukelberger: In offline discussion with @cgranade, @alan-geller, @martinquantum, @cpalmer2020, we agreed that (please comment if I misremember):

1. The operation should be named according to its effect rather than its machine-specific implementation, and hence `PermutationOracle` is more appropriate.
2. A generic, machine-agnostic implementation of `PermutationOracle` would be very nice, but is not a precondition for accepting the machine-specific (emulated) implementation. In general, the minimum requirement for accepting a machine-specific operation is that a generic version would undoubtedly be implementable. In other words, operations with unphysical effects are not acceptable, but it is clear that permutation oracles can be implemented for any machine (see synthesis methods as mentioned before).
3. As long as there is no generic implementation, the operation should not be `intrinsic`, but rather have a body with a `fail` statement. This makes it more apparent where an implementation must be provided when executing on a target machine that does not provide its own implementation.

The corresponding changes have been pushed.

Samples/src/OracleEmulation/Operations.qs
jgukelberger: Good ideas, done!

Samples/src/OracleEmulation/Operations.qs
jgukelberger: Agree, done!

Samples/src/OracleEmulation/Operations.qs
jgukelberger: Should be resolved as per discussion above.

Samples/src/OracleEmulation/Operations.qs
jgukelberger: Should be resolved as per discussion above.

jgukelberger:
@cgranade Any more comments or concerns? We should get some closure on this.

jgukelberger:
After discussing with @anpaz-msft, I've modified the definition of emulated oracles to align more closely with the scheme in [`EstimateFrequencyA`](https://github.com/microsoft/QuantumLibraries/blob/master/Standard/src/Characterization/EstimateFrequency.cs).

@anpaz-msft, @cgranade, what do you think?

Samples/src/OracleEmulation/PermutationOracle.cs
```diff
@@ -0,0 +1,222 @@
...
+        public static void ApplyOracle(QuantumSimulator simulator, Func<Int64, Int64, Int64> oracle,
+             IQArray<Qubit> xbits, IQArray<Qubit> ybits, bool adjoint = false)
+        {
+            var permutation = BuildPermutationTable(oracle, (int)xbits.Length, (int)ybits.Length);
+            ApplyOracle(simulator, permutation, xbits, ybits, adjoint);
+        }
+
+        /// <summary>
+        /// Apply a permutation defined by a permutation table. This overload
+        /// allows for perfomance optimizations like reuse of permutation
```
alan-geller: I think that caching the permutation table is an important optimization, and I'd be inclined to add that functionality either here or to the PermutationOracleImpl class. This could be done as a follow-on PR, but I think it's important to do it.

cgranade:
Thanks for the feedback, @alan-geller, and my apologies @jgukelberger for losing the notification on your original comments.

Samples/src/OracleEmulation/Driver.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public static void Pause()
+        {
+            System.Console.WriteLine("\n\nPress any key to continue...\n\n");
+            System.Console.ReadKey();
+        }
+
+        static void Main(string[] args)
+        {
+            // We begin by defining a quantum emulator (defined in Emulator.cs)
```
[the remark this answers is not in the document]
jgukelberger: Fixed, thanks!

Samples/src/OracleEmulation/Operations.qs
```diff
@@ -0,0 +1,169 @@
...
+        // `result`. Since all the oracles here ignore the flag, its length and
+        // state do not matter.
+        using ((flag, result) = (Qubit(), Qubit[8])) {
+            H(flag);
+
+            // Apply an oracle that was passed explicitly by the C# driver.
+            oracle([flag], result);
+            MeasureAndDisplay("The answer is ", result);
+
+            // Apply an oracle that was declared as `intrinsic` above and
```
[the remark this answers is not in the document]
jgukelberger: Fixed, thanks!

Samples/src/OracleEmulation/PermutationOracle.cs
jgukelberger: Sounds good. I'll defer that to a new PR because this one has been sitting around for a long time already.

PR #179, opened 2019-04-23
Going back to passing QArray into Run.

cgranade:


cgranade:
The CI should be restored soon, merging for now since it works locally.

PR #182, opened 2019-04-25
Adapt samples to Microsoft/QuantumLibraries#96.

cgranade:


Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -322,26 +320,26 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
...
-            
+
             for (i in 0 .. Length(gates) - 1) {
                 let (controls, target) = (gates[i])!;
                 Controlled X(controls, target);
             }
         }
-        
+
         adjoint invert;
```
bettinaheim: Do you want to remove those in favor of a declaration annotation?

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -354,33 +352,32 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
     /// # Output
     /// True, if the circuit correctly implements the permutation.
     operation PermutationSimulation (perm : Int[]) : Bool {
-        
+
         mutable result = true;
         let nbits = BitSizeI(Length(perm));
-        
         for (i in 0 .. Length(perm) - 1) {
```
bettinaheim: Looks like an IndexRange?

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -93,18 +93,18 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
     /// Sequence(23, 29); // [23, 24, 25, 26, 27, 28, 29]
     /// ```
     function Sequence (from : Int, to : Int) : Int[] {
```
bettinaheim: We should probably have a good initialization function in the Arrays module...
On one hand we need "conversions" between arrays and ranges (and have a conversation on related type system topics), on the other hand we should also have array initializers taking initialization functions.

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -122,17 +122,17 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
     /// Numbers(5); // [0, 1, 2, 3, 4]
     /// ```
     function Numbers (count : Int) : Int[] {
-        
+
         mutable array = new Int[count];
-        
+
         for (i in 0 .. count - 1) {
             set array w/= i <- i;
```
bettinaheim: Same here - essentially I'd like to get rid of these, because we should have convenient infrastructure in the array module.

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -154,105 +154,103 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
...
-    
+
+
     /// # Summary
     /// Constructs a MCMTMask type as a singleton array if targets is not 0,
     /// otherwise returns an empty array.
     function GateMask (controls : Int, targets : Int) : MCMTMask[] {
-        
+
         if (targets != 0) {
```
bettinaheim: return a conditional expression instead?

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -154,105 +154,103 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
...
+
     /// # Summary
     /// Update an output pattern according to gate mask.
     function UpdateOutputPattern (pattern : Int, gateMask : MCMTMask) : Int {
-        
+
         let (controls, targets) = gateMask!;
-        
+
         if ((pattern &&& controls) == controls) {
```
bettinaheim: return a conditional expression instead?

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -154,105 +154,103 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
...
         for (i in 0 .. Length(masks) - 1) {
             let (controls, targets) = (masks[i])!;
             let controlBits = IntegerBits(controls, n);
             let targetBits = IntegerBits(targets, n);
             let cQubits = Subarray(controlBits, qubits);
             let tQubits = Subarray(targetBits, qubits);
-            
+
             for (t in 0 .. Length(tQubits) - 1) {
                 set result = result + [MCTGate(cQubits, tQubits[t])];
```
bettinaheim: could be +=

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -389,12 +386,11 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
     /// An array of an even number of qubits.  The function pairs qubits from
     /// the first half of the array with the second half of the array.
     operation InnerProduct (qubits : Qubit[]) : Unit {
-        
 		let m = Length(qubits) / 2;
```
bettinaheim: there seem to be some tabs left here

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
```diff
@@ -407,17 +403,17 @@ namespace Microsoft.Quantum.Samples.ReversibleLogicSynthesis {
...
-        
+
         body (...) {
             let n = Length(qubits);
             let bits = IntAsBoolArray(shift, n);
             ApplyPauliFromBitString(PauliX, true, bits, qubits);
         }
-        
+
         adjoint invert;
```
bettinaheim: do you want to get rid of this in favor of a declaration annotation?

Samples/src/ReversibleLogicSynthesis/ReversibleLogicSynthesis.qs
cgranade: Very much agreed, let's open an issue for that on QuantumLibraries, then.

cgranade:
Thanks for the feedback, @bettinaheim! I've addressed your comments other than those relating to improved array initialization, which is now captured in the new issue Microsoft/QuantumLibraries#104.

PR #201, opened 2019-06-18
Added more GitHub metadata files.

cgranade:
This PR adds a new contribution guide and issue template for the Quantum Development Kit samples.


kant2002:
Incorrect link to Stack Overflow. It point to vscode tag, instead of qsharp

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
+# Contributing to the Quantum Development Kit Samples #
+
+Welcome, and thank you for your interest in contributing to the Quantum Development Kit!
+
+There are many ways in which you can contribute, beyond writing code. The goal of this document is to provide a high-level overview of how you can get involved.
+For more details on how to contribute to the samples or to the rest of the Quantum Development Kit, please see the [contribution guide](https://docs.microsoft.com/quantum/contributing/).
+
+## Asking Questions
+
+Have a question? Rather than opening an issue, please ask away on [Stack Overflow](https://stackoverflow.com/questions/tagged/vscode) using the tag `qsharp`.
```
tcNickolas: @kant2002's suggestion

```suggestion
Have a question? Rather than opening an issue, please ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/q%23) or [Quantum Computing StackExchange](https://quantumcomputing.stackexchange.com/questions/tagged/q%23) using the tag `q#`.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+The active community will be eager to assist you.
```
tcNickolas: I feel that at this point of time the Q# community on SO consists mostly of our teammates. At this point I'd stress ease of search (the sentence you have on the second line) rather than the active community.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+* Your operating system
+
+* Reproducible steps (1... 2... 3...) that cause the issue
+
+* What you expected to happen, versus what actually happened
+
+## Contributing Documentation
+
+If you are interested in contributing to conceptual documentation about the Quantum Development Kit, please see the [MicrosoftDocs/quantum-docs-pr](https://github.com/MicrosoftDocs/quantum-docs-pr) repository.
+If you are interested in contributing API references, please see [Contributing Code](https://docs.microsoft.com/quantum/contributing/code) in the contribution guide.
```
tcNickolas:
```suggestion
If you are interested in contributing to API references, please see [Contributing to the API References](https://docs.microsoft.com/quantum/contributing/docs#contributing-to-the-api-references) in the contribution guide.
```
?

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Contributing Fixes and New Features
+
+If you are interested in writing code to fix issues or to implement new features, please see [Contributing Code](https://docs.microsoft.com/quantum/contributing/code) in the contribution guide.
```
tcNickolas: Since this is specifically samples repo, would it make sense to say "new samples" instead? "Features" makes me think more about language or library features

CONTRIBUTING.md
tcNickolas: I'm also not sure we should send people with all questions to SO exclusively. SO's FAQ has a [question regarding this](https://meta.stackoverflow.com/questions/253849/is-it-acceptable-to-use-stack-overflow-as-a-qa-for-a-specific-product), which says explicitly "**Make sure we're not your only support option** - and let folks pick which one works the best for them. Make sure they can easily find a place to just ask you their question on the same page that points them to us." Following this guidance, we can suggest asking the question on SO but we should not discourage asking questions directly in the repo - or direct people to another channel that allows to reach the team directly.

CONTRIBUTING.md
cgranade: Fair point, agreed. Was trying to stick to the general format that has been used in other MSFT repos, but you're right that it is a bit confusing here. Will reword.

CONTRIBUTING.md
cgranade: > I feel that at this point of time the Q# community on SO consists mostly of our teammates. At this point I'd stress ease of search (the sentence you have on the second line) rather than the active community.

I think that the Q# community is growing pretty substantially, both on and off of StackOverflow such that pointing to that community still makes sense here.



> I'm also not sure we should send people with all questions to SO exclusively. SO's FAQ has a [question regarding this](https://meta.stackoverflow.com/questions/253849/is-it-acceptable-to-use-stack-overflow-as-a-qa-for-a-specific-product), which says explicitly "**Make sure we're not your only support option** - and let folks pick which one works the best for them. Make sure they can easily find a place to just ask you their question on the same page that points them to us." Following this guidance, we can suggest asking the question on SO but we should not discourage asking questions directly in the repo - or direct people to another channel that allows to reach the team directly.

I take the point, but I was following a pretty standard Microsoft template here. I'm happy to point to other support channels, but I'm not sure what the best channels would be in this case. I'm a bit reticent to suggest that people file questions as GitHub Issues, since that somewhat conflates with feature requests and bug reports. @cpalmer2020, do you have any suggestions here? Thanks!

cpalmer2020: I see that the proposed language is the same as followed by VS Code, and I believe that that language follows the guidance of the FAQ.  We provide a pointer to StackOverflow from the contribution Guide, and at that same place in our guide, we provide users with information on how to file an issue.  Thus users are aware of the option of filing an issue.  If we do get questions filed as issues, we will directly answer them, so we are following the FAQ.

tcNickolas: I read "Rather than opening an issue, please ask away" as actively discouraging people from asking questions in issues, so I felt I should point this out. You do not have to change the language to address this - this is why I "left review comments" instead of "requested changes".

CONTRIBUTING.md
```diff
@@ -7,10 +7,9 @@ For more details on how to contribute to the samples or to the rest of the Quant
 
 ## Asking Questions
 
-Have a question? Rather than opening an issue, please ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/q%23) or [Quantum Computing StackExchange](https://quantumcomputing.stackexchange.com/questions/tagged/q%23) using the tag `q#`.
-
-The active community will be eager to assist you.
-Your well-worded question will serve as a resource to others searching for help.
+Have a question? We have already answered it on the [Q# Development Blog](https://devblogs.microsoft.com/qsharp/), if not feel free to ask us in a comment!
```
tcNickolas: Looks better, but the "we have already answered it" feels a bit odd to me... I would suggest the following:

Have a question? The `q#` tags on [Stack Overflow](https://stackoverflow.com/questions/tagged/q%23) and [Quantum Computing StackExchange](https://quantumcomputing.stackexchange.com/questions/tagged/q%23) are also great places to ask questions about Q#, Quantum Development Kit and quantum computing in general. You can learn more about our work on the [Q# Blog](https://devblogs.microsoft.com/qsharp/) and ask questions in the comments as well.

(This way we stress that quantum computing in general questions are better on SO, while still sounding welcoming, and point to blog without saying that we've foreseen your questions)

(And keep "However you ask it" sentence, I like it)

cgranade: That was supposed to read "we _may_ have already answered it," but apparently I can't type, sigh. That said, your wording is better anyway, so I've applied that. Thanks!

PR #202, opened 2019-06-19
Begin simplifying and updating the Get-GateCount sample

cgranade:
This PR begins to simplify the gate count sample by splitting the one Program.cs into several files that are each easier to read, making outputting CSVs optional, and by consolidating common logic into a few extension methods.


Chemistry/GetGateCount/3-GetGateCount.csproj
```diff
@@ -16,9 +16,9 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.1.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.1.1" />
     <PackageReference Include="Microsoft.PowerShell.5.ReferenceAssemblies" Version="1.1.0" NoWarn="NU1701" />
-    <PackageReference Include="Microsoft.Quantum.Chemistry" Version="0.7.1905.3109" />
-    <PackageReference Include="Microsoft.Quantum.Standard" Version="0.7.1905.3109" />
-    <PackageReference Include="Microsoft.Quantum.Development.Kit" Version="0.7.1905.3109" />
+    <PackageReference Include="Microsoft.Quantum.Chemistry" Version="0.8.1906.1704-beta" />
```
anpaz-msft: should we merge this into master?

Chemistry/GetGateCount/Configuration.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// <summary>
+    /// Enumeration type for choice of Hamiltonian simulation algorithm
+    /// </summary>
+    public enum HamiltonianSimulationAlgorithm
+    {
+        ProductFormula,
+        Qubitization
+    };
+    /// <summary>
```
anpaz-msft: I would leave an extra blank line for readability.

Chemistry/GetGateCount/Configuration.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            Order = setOrder;
+            if (setOrder > 2)
+            {
+                throw new System.NotImplementedException($"Product formulas of order > 2 not implemented.");
+            }
+            StepSize = setStepSize;
+        }
+    };
+    /// <summary>
```
anpaz-msft: I would leave an extra blank line for readability.

Chemistry/GetGateCount/Configuration.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// Configuration data for Qubitization simulation algorithm
+    /// </summary>
+    public struct QubitizationConfig
+    {
+        /// <summary>
+        /// Choice of quantum state preparation
+        /// </summary>
+        public QubitizationStatePrep qubitizationStatePrep;
```
anpaz-msft: should this start with capital Q?

Chemistry/GetGateCount/Configuration.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// <summary>
+    /// Configuration data for Hamiltonian simulation algorithm
+    /// </summary>
+    public struct HamiltonianSimulationConfig
+    {
+        /// <summary>
+        /// Choice of Hamiltonian simulation algorithm
+        /// </summary>
+        public HamiltonianSimulationAlgorithm hamiltonianSimulationAlgorithm;
```
anpaz-msft: These should be changed to Properties.

Chemistry/GetGateCount/GetGateCount.cs
```diff
@@ -0,0 +1,270 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// This loads a Hamiltonian from file and performs gate estimates of a
+// - Jordan–Wigner Trotter step
+// - Jordan–Wigner Qubitization iterate
+
+#region Using Statements
+// We will need several different libraries in this sample.
```
anpaz-msft:
```suggestion
// We will use several different libraries in this sample.
```

Chemistry/GetGateCount/GetGateCount.cs
```diff
@@ -0,0 +1,270 @@
...
+// Stopwatch class, which is quite useful for measuring
+// how long each gate counting run takes.
+using System.Diagnostics;
+
+// The System.Collections.Generic library provides many different
+// utilities for working with collections such as lists and dictionaries.
+using System.Collections.Generic;
+
+
+// Finally, we use the Mono.Options and System.Management.Automation
```
anpaz-msft: These are not the libraries listed below :)

Chemistry/GetGateCount/GetGateCount.cs
```diff
@@ -0,0 +1,270 @@
...
+    {
+        public string IntegralDataPath;
+        public string Method;
+        public double SpinOrbitals;
+        public double TCount;
+        public double RotationsCount;
+        public double CNOTCount;
+        public double NormMultipler;
+        public long ElapsedMilliseconds;
+        public Dictionary<string, string> TraceSimulationStats;
```
anpaz-msft: Should we keep these? I don't see them used or printed.

Chemistry/GetGateCount/PowerShellIntegration.cs
```diff
@@ -6,32 +6,20 @@
...
-// The System.Diagnostics namespace provides us with the
-// Stopwatch class, which is quite useful for measuring
-// how long each gate counting run takes.
-using System.Diagnostics;
-
 // The System.Collections.Generic library provides many different
 // utilities for working with collections such as lists and dictionaries.
 using System.Collections.Generic;
 
 // We use the logging library provided with .NET Core to handle output
```
anpaz-msft: Looks like this is gone.

Chemistry/GetGateCount/PowerShellIntegration.cs
```diff
@@ -6,32 +6,20 @@
...
 // in a robust way that makes it easy to turn on and off different messages.
-using Microsoft.Extensions.Logging;
 
-// Finally, we use the Mono.Options and System.Management.Automation
-// libraries to make it easy to use this sample from the command line.
-using Mono.Options;
+// We use the System.Management.Automation
+// library to integrate with PowerShell.
 using System.Management.Automation;
+using System.Linq;
```
anpaz-msft: No comment for System.Linq.

Chemistry/GetGateCount/GetGateCount.cs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            // Creates and configures Trace simulator for accumulating gate counts.
+            var sim = CreateAndConfigureTraceSim();
+
+            // Create stop-watch to measure the execution time
+            var stopWatch = new Stopwatch();
+
+            var gateCountResults = new GateCountResults();
+            #region Trace Simulator on Trotter step
+            if (config.hamiltonianSimulationAlgorithm == HamiltonianSimulationAlgorithm.ProductFormula)
```
anpaz-msft: Each of these are mostly copy/paste... I would take the Configuration concept and go a step further and move the which Q# operation to run as part of the configuration and a `Name` property, so you don't have to have the code duplication here.

Chemistry/GetGateCount/Configuration.cs
```diff
@@ -0,0 +1,118 @@
...
+    public enum QubitizationStatePrep
+    {
+        MinimizeQubitCount,
+        MinimizeTGateCount
+    }
+
+    /// <summary>
+    /// Configuration data for Hamiltonian simulation algorithm
+    /// </summary>
+    public struct HamiltonianSimulationConfig
```
anpaz-msft: Instead of a `HamiltonianSimulationConfig` class with different config properties depending on the type of algorithm, I would create an `IHamiltonianSimulationConfig` interface with a couple of methods needed to run the simulation, for example `HamiltonianSimulationAlgorithm` and `EntryOperation`, and then have the `ProductFormulaConfig` and `QubitizationConfig` implement this interface.
This would also address the duplicated code in `RunGateCount`.

Chemistry/GetGateCount/3-GetGateCount.csproj
cgranade: Ah, good point. I'll target a new release/v0.8 branch, then.

Chemistry/GetGateCount/Configuration.cs
cgranade: Good point, added those.

Chemistry/GetGateCount/Configuration.cs
cgranade: Good point, changing.

Chemistry/GetGateCount/Configuration.cs
cgranade: Refactoring accordingly.

Chemistry/GetGateCount/PowerShellIntegration.cs
cgranade: Ah, good catch. Thanks!

Chemistry/GetGateCount/PowerShellIntegration.cs
cgranade: Added.

Chemistry/GetGateCount/GetGateCount.cs
cgranade: They are in principle accessible via the PowerShell interface, but they're not otherwise used, no.

Chemistry/GetGateCount/GetGateCount.cs
cgranade: Agreed. I actually tried to do that, but ran into a couple slight issues in that the `GetMetric` method of the trace simulator takes a type argument for the operation in question, and in that the optimized qubitization operation takes an additional parameter for the target error. I'd prefer if possible to defer this improvement to the next PR on this sample if that's OK.

cgranade:
Merging in now, with future work tracked in #203. Thanks!

PR #206, opened 2019-07-10
Update samples for 0.8 beta.

cgranade:


.devcontainer/README.md
```diff
@@ -0,0 +1,121 @@
+# Remote Development Container for Visual Studio Code (preview) #
+
+This folder defines a _development container_ for the Quantum Development Kit samples to make it easier to get up and running with trying out Q# with C#, Python, and Jupyter Notebook.
+
+## What is a Development Container? ##
+
+Visual Studio Code allows for using [Docker](https://docs.microsoft.com/dotnet/standard/microservices-architecture/container-docker-introduction/docker-defined) to quickly define development environments, including all the compilers, command-line tools, libraries, and programming platforms you need to get up and running quickly.
+Using the definitions provided in this folder, Visual Studio Code can use Docker automatically install the correct version of the Quantum Development Kit as well as other software you might want to use with Q#, such as Python and Jupyter Notebook --- all into an isolated container that doesn't affect the rest of the software you have on your system.
```
anpaz-msft:
```suggestion
Using the definitions provided in this folder, Visual Studio Code can use Docker to automatically install the correct version of the Quantum Development Kit as well as other software you might want to use with Q#, such as Python and Jupyter Notebook --- all into an isolated container that doesn't affect the rest of the software you have on your system.
```

Samples/src/Arithmetic/AdderExample.ipynb
```diff
@@ -235,15 +235,15 @@
    "outputs": [
     {
      "data": {
```
anpaz-msft: is it better to commit the notebooks with no output?

cgranade:
> Somehow the FSharp seems to broke the build.

That would explain why it only broke the Linux build... let me go on and fix that, then. Thanks!

Samples/src/Arithmetic/AdderExample.ipynb
cgranade: I could see either way, really. Including the output allows for the notebooks to be read on GitHub and nbviewer, but does make the files much larger, harder to maintain, etc.

PR #218, opened 2019-08-08
Updated quantum teleportation Jupyter notebook to use more precise language

cesarzc:
Changed the verb copy to transfer to describe the teleport operation so that it is not confusing given that the non-cloning theorem states that a qubit cannot be copied the same way a classical bit can.

I also added a note and a link to Microsoft's documentation about the non-cloning theorem.


Samples/src/Teleportation/Notebook.ipynb
```diff
@@ -83,6 +83,13 @@
     "}\n"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that the state of the original qubit is destroyed when we perform a measurement on it, so the `Teleport` Q# operation does not violate the [non-cloning theorem](https://docs.microsoft.com/en-us/quantum/concepts/pauli-measurements?view=qsharp-preview#the-no-cloning-theorem).\n"
```
cgranade:
```suggestion
    "Note that the state of the original qubit is destroyed when we perform a measurement on it, so the `Teleport` Q# operation does not violate the [non-cloning theorem](https://docs.microsoft.com/quantum/concepts/pauli-measurements#the-no-cloning-theorem).\n"
```

cesarzc:
@cgranade, what is the process to merge the pull request?

cgranade:
My bad, @cesarzc, I forgot to hit the button after the tests completed. Will get that merged in. Thanks again for the fix!

cesarzc:
@cgranade thank you for reviewing and merging the PR!

PR #221, opened 2019-08-12
Add a Q# quantum state visualizer

marshallsa:
This is a web-based debugger that can step (forward and back) through a quantum simulation, showing the operation call stack and quantum state at each step. It's based on a visualizer by Chris Granade.

Co-authored-by: Chris Granade <chgranad@microsoft.com>


anpaz-msft:
Looks pretty cool! Please do give me the change to review it. I'm on vacation but plan to spend some time on Thursday.
Thanks!

Samples/src/VisualDebugger/Hello.qs
```diff
@@ -0,0 +1,60 @@
+namespace vis_sim {
```
anpaz-msft: Missing copyright notice.

Samples/src/VisualDebugger/Hubs.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.AspNetCore.SignalR;
+using System.Threading.Tasks;
+
+namespace vis_sim
```
anpaz-msft: We should probably put this sample under the `Microsoft.Quantum.Samples.VisualDebugger` namespace...

Samples/src/VisualDebugger/Hubs.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.AspNetCore.SignalR;
```
anpaz-msft: copyright notice.

Samples/src/VisualDebugger/Hubs.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Threading.Tasks;
+
+namespace vis_sim
+{
+    internal class VisualHub : Hub
```
anpaz-msft: Let's add some ///summary with what this class does and how it works, for those not familiar with SignalR.

Samples/src/VisualDebugger/Program.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.Quantum.Simulation.Simulators;
```
anpaz-msft: Copyright notice.

Samples/src/VisualDebugger/Program.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace vis_sim
```
anpaz-msft: Namespace

Samples/src/VisualDebugger/Program.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    internal class Program
```
anpaz-msft: Similarly, let's add some description about what this program does and how to call it from the command line.
A couple of troubleshooting tips would be nice (although those can go in the README.md)

Samples/src/VisualDebugger/Startup.cs
```diff
@@ -0,0 +1,26 @@
+using Microsoft.AspNetCore.Builder;
```
anpaz-msft: Copyright notice.

Samples/src/VisualDebugger/Startup.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace vis_sim
```
anpaz-msft: Namespace

Samples/src/VisualDebugger/Startup.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal class Startup
```
anpaz-msft: At first glance, it's unclear about why there is a `Startup` and why it's not called from `Program.Main`. Adding some description here would be nice.

Samples/src/VisualDebugger/VisualDebugger.cs
```diff
@@ -0,0 +1,118 @@
...
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace vis_sim
```
anpaz-msft: Namespace.

Samples/src/VisualDebugger/VisualDebugger.cs
```diff
@@ -0,0 +1,118 @@
...
+        private readonly QuantumSimulator simulator;
+        private readonly StateDumper stateDumper;
+        private readonly IWebHost host;
+        private readonly IHubContext<VisualHub> context;
+        private readonly ManualResetEvent advanceEvent = new ManualResetEvent(true);
+        private readonly IList<(string method, object[] args)> history = new List<(string, object[])>();
+
+        public VisualDebugger(QuantumSimulator simulator)
+        {
+            simulator.OnOperationStart += OnOperationStartHandler;
```
anpaz-msft: You should check that simulator != null and throw `ArgumentNullException` at top.

Samples/src/VisualDebugger/VisualDebugger.cs
```diff
@@ -0,0 +1,118 @@
...
+        public async Task Run(Func<IOperationFactory, Task<QVoid>> operation)
+        {
+            await operation(simulator);
+        }
+
+        public bool Advance() => advanceEvent.Set();
+
+        public async Task ReplayHistory(IClientProxy client)
+        {
+            foreach (var (method, args) in history)
```
anpaz-msft: Can you please add brackets to wrap the foreach?

Samples/src/VisualDebugger/VisualDebugger.cs
```diff
@@ -0,0 +1,118 @@
...
+                await client.SendCoreAsync(method, args);
+        }
+
+        private T GetService<T>() => (T) host.Services.GetService(typeof(T));
```
anpaz-msft: By internal convention, the body after `=>` should be on the next line.

Samples/src/VisualDebugger/VisualDebugger.cs
```diff
@@ -0,0 +1,118 @@
...
+            await context.Clients.All.SendCoreAsync(method, args);
+        }
+
+        private async Task WaitForAdvance() => await Task.Run(() =>
+        {
+            advanceEvent.Reset();
+            advanceEvent.WaitOne();
+        });
+
+        private void OnOperationStartHandler(ICallable operation, IApplyData arguments)
```
anpaz-msft: Chris added recently the `OnQubitsAllocation` (or similar) to the `SimulatorBase`. You should change the code here to leverage it.

Samples/src/VisualDebugger/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
anpaz-msft: I'm under the impression that `package-lock.json` should not be committed, but I could be wrong.

Samples/src/VisualDebugger/webpack.config.js
```diff
@@ -0,0 +1,37 @@
+const path = require("path");
```
anpaz-msft: I'm not convinced that using `webpack` give us any benefit. We literally have 3 files (one .ts, one .html and one .css), adding an extra build step to pack this seems too much work for what we need.

Samples/src/VisualDebugger/websrc/app.ts
```diff
@@ -0,0 +1,304 @@
...
+
+const stateChart = new chart.Chart(chartContext, {
+    type: "bar",
+    data: {
+        labels: [],
+        datasets: [
+            {
+                data: [],
+                label: "Real",
+				backgroundColor: "#ff0000",
```
anpaz-msft: The file has some tabs , please convert all to spaces.

Samples/src/VisualDebugger/websrc/app.ts
```diff
@@ -0,0 +1,304 @@
...
+                        suggestedMin: -1,
+                        suggestedMax: 1
+                    }
+                }
+            ]
+        }
+    }
+});
+
+function updateChart(state: State) {
```
anpaz-msft: It would be nice if the x-axis showed the base vector as the label, is that possible?

marshallsa: Yes, they should be there - are they not working? This is what I see:

Basis Vectors

Samples/src/VisualDebugger/package.json
marshallsa: It looks like the npm documentation says it's intended to be committed as a snapshot of all the package versions in node_modules: https://docs.npmjs.com/files/package-lock.json

Samples/src/VisualDebugger/webpack.config.js
marshallsa: I think we need something like Webpack or Browserify to make the npm modules we import work in the browser (`@aspnet/signalr` and `chart.js`). I tried compiling directly with `tsc` and it didn't work.

Samples/src/VisualDebugger/Program.cs
marshallsa: Did you run into any issues that you think should be mentioned in a troubleshooting section?

marshallsa:
@anpaz-msft Thanks for your review! I think I've addressed all of your comments (except for a few questions I had). Could you take another look when you have time?

Samples/src/VisualDebugger/websrc/app.ts
anpaz-msft: Ah! You're right. I was expecting them to be in the x-axis line (i.e. on the `0`); further more, they actually do show but they are mostly covered by the scrollbar:

image

It's ok if we leave them at the bottom, but could we push them a little higher so they are not blocked by the scrollbar?

Further, allocating new qubits is not showing the new basis state after the allocation, i.e.:
image

Is that something we could add?

anpaz-msft:
One more comment: calling this a Visual Debugger is a bit of a stretch as you can't do most of the tasks normally associated with a debugger (check variables, put breakpoints, evaluate expressions, etc). Maybe we should call it "QuantumStateVisualizer" or just "StateVisualizer" instead?

Samples/src/VisualDebugger/websrc/app.ts
marshallsa: Hmm, I don't see the bottom scrollbar even when I make my window small. What browser are you using?

marshallsa: It seems like the `OnAllocateQubits` handler is called just before the qubits are actually allocated - so the state that I can get in the handler is out of date. For example after the third allocation, the state would show only 4 basis vectors, but after the first H it would show 8 vectors.

@cgranade Is it possible for the simulator to invoke `OnAllocateQubits` after calling `manager.Allocate`?

anpaz-msft: @marshallsa, OnOperationStart is called *before* the operation is called, `onAllocate` should remain consistent with that.
Are you saying that before calling 'H', there are only 4 basis state? that seems odd, there might be something else going on, by the time 'H' is highlighted (i.e. after `Allocate` is called) there should already be 8 basis state vectors.

marshallsa: No, I think you're right that there are 8 basis vectors by the time `OnOperationStart` is called for H, but I was trying to show all 8 of them on the last "allocate 1 qubit" line instead of the first "H" line. Do you prefer showing 4 basis states on the last allocate line and then at the start of the first H line (before H actually runs), it shows 8?

anpaz-msft: I think it's ok to be consistent, so when the last Allocation is highlighted then there should be 4, when H gets highlighted there should be 8. Again, the instruction listed is what is going to be executed, not what was already executed.

anpaz-msft:
You're going to hate me, but for some reason the dotnet process keeps ligering around once you stop using Ctrl+C. My experience is that this is typically that you're starting a thread and continue execution on Main, Main finishes but execution continues due to the open thread, but its' just an idea.

marshallsa:
@anpaz-msft Ctrl+C should work now. :)

Do you still have the bug where the scrollbar at the bottom covers up the basis vector labels? I haven't been able to reproduce that.

Samples/src/VisualDebugger/websrc/app.ts
marshallsa: That makes sense - allocation should work that way now. I also changed the green checkmark to a blue arrow like the rest of the operations so it's clear that the allocation is the next thing to execute. Sorry for the confusion!

Samples/src/StateVisualizer/websrc/css/main.css
```diff
@@ -0,0 +1,83 @@
...
+
+*, *::before, *::after {
+    box-sizing: border-box;
+}
+
+html, body {
+    margin: 0;
+    padding: 0;
+    height: 100%;
+    width: 100%;
```
anpaz-msft:
```suggestion
    width: 100%;
    overflow: hidden;
```

anpaz-msft: this fixes the superfluos scrollbar that is hidden the base states.

Samples/src/StateVisualizer/package-lock.json
```diff
@@ -0,0 +1,5366 @@
+{
+  "name": "vis-sim",
```
anpaz-msft:
```suggestion
  "name": "qs-state-visualizer",
```

anpaz-msft: that's what it generated on my laptop... again, I don't think this should be committed :)

marshallsa: Oops, I forgot to update package-lock.json after renaming the package. If you want to add it to gitignore I am also OK with that. :)

marshallsa:
> One very minor thing.
> Should I merge it after it?

Sure, I think it should be ready to merge (unless you want to remove package-lock.json first).

anpaz-msft:
Thanks for incorporating all the feedback. Great contribution!

PR #224, opened 2019-09-04
Add a very basic implementation of Grover's algorithm

thomashaener:
This PR adds a new sample that shows how to implement Grover's algorithm in Q#. The sample is supposed to be as simple as possible (e.g., for use in beginner tutorials).
The marking oracle uses a bit-flip oracle (as opposed to using a phase-flip directly) to better align with textbook descriptions and to make it easier to experiment with. E.g., one could combine it with the OracleSynthesis sample.


Samples/src/SimpleGrover/README.md
```diff
@@ -0,0 +1,14 @@
+﻿# Grover Search Sample #
```
msoeken: You could link to the more extended DatabaseSearch example in the README.

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Samples.SimpleGrover {
```
msoeken: I suggest to add a new line after the copyright header.

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Convert;
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Arrays;
+    open Microsoft.Quantum.Arithmetic;
+
+    // This operation adds a (-1)-phase to the marked state(s)
+    operation ReflectMarked (inputQubits : Qubit[]) : Unit is Adj + Ctl {
```
msoeken: You never call the adjoint or a controlled version of the function, so `Adj + Ctl` could be removed.

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+                // Initialize the outputQubit to 1/sqrt(2) ( |0> - |1> )
+                // so that toggling it results in a (-1)-phase
+                X(outputQubit);
+                H(outputQubit);
+            }
+            apply {
+                // Flip the outputQubit for marked states.
+                // Here: For the state with alternating 0s and 1s
+                within {
+                    ApplyToEachCA(X, inputQubits[0..2..Length(inputQubits)-1]);
```
msoeken: Considering to drop `Adj + Ctl` from this operation, this can become `ApplyToEach`

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+                }
+                apply {
+                    (Controlled X) (inputQubits, outputQubit);
+                }
+            }
+        }
+    }
+    
+    // This operation adds a (-1)-phase to the uniform superposition
+    operation ReflectUniform (inputQubits : Qubit[]) : Unit is Adj + Ctl {
```
msoeken: See above.

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+                }
+            }
+        }
+    }
+    
+    // This operation adds a (-1)-phase to the uniform superposition
+    operation ReflectUniform (inputQubits : Qubit[]) : Unit is Adj + Ctl {    
+        within {
+            // Transform the uniform superposition to all-zero
+            ApplyToEachCA(H, inputQubits);
```
msoeken: This could be simplified to `ApplyToEachA`.

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+            for (i in 0..numIterations-1) {
+                ReflectMarked(qubits);
+                ReflectUniform(qubits);
+            }
+            // measure the answer
+            let measuredInt = MeasureInteger(LittleEndian(qubits));
+            return IntAsBoolArray(measuredInt, numQubits);
+        }
+    }
+}
```
msoeken: Please add an empty line to the end of the file.

Samples/src/SimpleGrover/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This sample implements Grover's search algorithm. Oracles implementing the database are explicitly constructed together with all steps of the algorithm.
+
+## Running the Sample in Visual Studio ##
+
+Open the `QsharpSamples.sln` solution in Visual Studio and set the .csproj file in the manifest as the startup project.
+Press Start in Visual Studio to run the sample.
```
cgranade: I'd suggest adding a `host.py` along with instructions on how to run the sample from the command line using either Python or the .NET Core SDK:

````suggestion
Press Start in Visual Studio to run the sample.

## Running the Sample from the Command Line ##

This sample can also be run from the command line, including from the integrated terminal in Visual Studio Code.
To run the C# host program for this sample, use the `dotnet` command:

```bash
dotnet run
```

To run this sample from Python:

```bash
python host.py
```
````

File: Samples/src/SimpleGrover/SimpleGrover.qs (hunk 4 above)
Comment (cgranade) [reply Samples/src/SimpleGrover/SimpleGrover.qs]: Support for `Adjoint` is required for operations called inside a `within` block, but I agree that controlled shouldn't be required. It may be at the moment, though, as I'm not sure that https://github.com/microsoft/qsharp-compiler/pull/152 is included in the 0.9 release.

File: Samples/src/SimpleGrover/SimpleGrover.qs [hunk 9]
```diff
@@ -0,0 +1,69 @@
...
+            within {
+                // Initialize the outputQubit to 1/sqrt(2) ( |0> - |1> )
+                // so that toggling it results in a (-1)-phase
+                X(outputQubit);
+                H(outputQubit);
+            }
+            apply {
+                // Flip the outputQubit for marked states.
+                // Here: For the state with alternating 0s and 1s
+                within {
```
cgranade: Would it make sense to remove a level of nesting here?

Samples/src/SimpleGrover/SimpleGrover.qs
cgranade:
```suggestion
    operation ReflectAboutMarked(inputQubits : Qubit[]) : Unit is Adj + Ctl {
```

Samples/src/SimpleGrover/SimpleGrover.qs
cgranade:
```suggestion
    operation ReflectAboutUniform(inputQubits : Qubit[]) : Unit is Adj + Ctl {    
```

Samples/src/SimpleGrover/SimpleGrover.qs
```diff
@@ -0,0 +1,69 @@
...
+        using (qubits = Qubit[numQubits]) {
+            // initialize uniform superposition
+            ApplyToEach(H, qubits);
+            // perform iterations
+            for (i in 0..numIterations-1) {
+                ReflectMarked(qubits);
+                ReflectUniform(qubits);
+            }
+            // measure the answer
+            let measuredInt = MeasureInteger(LittleEndian(qubits));
```
cgranade: It feels a little funny to me to measure an integer, then immediately convert it to a `Bool[]`. Would it make sense to either return the `Int` directly, or to use `return ForEach(MResetZ, qubits)` here?

alan-geller:
How does this sample relate to the Grover Kata, https://github.com/microsoft/QuantumKatas/tree/master/GroversAlgorithm?

thomashaener:
> How does this sample relate to the Grover Kata, https://github.com/microsoft/QuantumKatas/tree/master/GroversAlgorithm?

In my opinion, the purpose of the Katas is to teach programmers (in this particular case how to implement Grover), whereas samples illustrate how concise complete applications can be because of the language features that Q# provides (such as, e.g., within-apply in this sample).

Samples/src/SimpleGrover/SimpleGrover.qs
thomashaener: The outer level turns the bit-flip oracle into a phase-flip oracle, and the inner within-apply is the bit-flip oracle itself, so I left the two levels for illustration. Should I put an additional note there or similar?

thomashaener:
@cgranade @msoeken Thanks a lot for the feedback! I've addressed all comments & suggestions in the newest commit.

Samples/src/SimpleGrover/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+        static void Main(string[] args)
+        {
+            var sim = new QuantumSimulator();
+            var nDatabaseQubits = 5;
+
+            var result = ApplyGrover.Run(sim, nDatabaseQubits).Result;
+            Console.WriteLine($"Result: {result}");
+
+            System.Console.WriteLine("\n\nPress any key to continue...\n\n");
+            System.Console.ReadKey();
```
msoeken: For consistency with line 17, I suggest to remove `System.` from the beginning of the line.

thomashaener: Good catch, thanks!

cgranade:
Thanks for adding this sample, @thomashaener!

PR #235, opened 2019-10-04
Reorganize samples to being onboarding to https://docs.microsoft.com/samples/.

cgranade:
This draft PR begins reorganizing samples to make the categorization more clear (e.g.: placing chemistry and numerics samples under the `samples/` folder), and to add metadata to "Getting Started" samples needed for inclusion in https://docs.microsoft.com/samples/.

**TODO**:
- [x] Make sure solution files for each folder under `samples/` exist and are correct.
- [x] Update README.md at the root, and in each folder under `samples/`.


samples/getting-started/intro-to-iqsharp/README.md
```diff
@@ -0,0 +1,28 @@
+---
+page_type: sample
+languages: [qsharp]
+products: [qdk]
+---
+
+# Intro to IQ# #
```
natke: I would suggest renaming this to something like: "Intro to Q# Jupyter notebooks"

samples/getting-started/teleportation/README.md
```diff
@@ -0,0 +1,61 @@
+﻿---
+page_type: sample
+languages: [qsharp, python, csharp]
+products: [qdk]
+---
+
+# Quantum Teleportation Sample #
```
natke: We don't need the pounds around the headings, just at the beginning. Having said that, I'm not sure if they will be rendered as # or ignored by the markdig renderer.

samples/getting-started/intro-to-iqsharp/README.md
cgranade: Good point, will go on and rename, thanks!

samples/getting-started/teleportation/README.md
cgranade: I always learned the style with `#` at both ends, but I'm fine either way; happy to normalize to only leading `#` if that's more consistent.

samples/getting-started/measurement/Host.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    class Program
+    {
+
+        static void Main(string[] args)
+        {
+
+            using (var qsim = new QuantumSimulator())
+            {
+                RunQuantumMain.Run(qsim).Wait();
```
bettinaheim: I am really looking forward to getting rid of the need for this!

PR #252, opened 2019-10-22
Update relative links to absolute in sample manifests.

cgranade:
This PR modifies sample manifests to be absolute, so that links render correctly at https://docs.microsoft.com/samples as well as in the repo itself. Thanks to @crazy4pi314 and @lobrien for reporting the issue!


cgranade:
Abandoning in favor of #264.

PR #280, opened 2019-12-16
Minor bug fixes for the state visualizer sample

marshallsa:
Just a few bug fixes for things I found while writing my Q# Advent Calendar post about the state visualizer:

1. There was something wrong with the serialization of operation return values (unit appeared in JavaScript as `{qubits: null}`, and `One` and `Zero` were both `{}` for some reason). I convert them to strings before sending them to the web client now.
2. Fixed a bug where log messages did not cause the timeline to scroll down when they were added (they were missing an element with the `operation-name` class).
3. Removed a period that was at the end of the page title. :)

### DIFF
--- a/samples/runtime/state-visualizer/StateVisualizer.cs
+++ b/samples/runtime/state-visualizer/StateVisualizer.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Quantum.Samples.StateVisualizer
 
         private void OnOperationEndHandler(ICallable operation, IApplyData result)
         {
-            BroadcastAsync("OperationEnded", result?.Value.ToString(), stateDumper.DumpAndGetAmplitudes()).Wait();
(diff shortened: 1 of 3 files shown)


bettinaheim:
Thank you, @marshallsa, for putting this up! And of course for writing a blog post. :) 
Looks good to me, @cgranade do you want to give it a second look?

cgranade:
Thanks for the bug fixes, @marshallsa, and for the post!

> 1. There was something wrong with the serialization of operation return values (unit appeared in JavaScript as `{qubits: null}`, and `One` and `Zero` were both `{}` for some reason). I convert them to strings before sending them to the web client now.

Makes sense. In the future, it would be good to have a more formal approach to serializing Q# values into JSON so that we can better support debuggers, but I agree with the approach of converting to strings on the server end until that's in place.

PR #282, opened 2019-12-17
Create new Docker publishing workflow.

cgranade:
This PR sets up a new GitHub Action to prebuild the Docker image for the samples repo and publish it to the packages feed for this repo.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,20 +17,20 @@ jobs:
       uses: Azure/docker-login@v1
       with:
         # Container registry username
-        username: microsoft
+        username: ${{ secrets.ACR_SERVICE_PRINCIPAL_NAME }}
         # Container registry password
-        password: ${{ secrets.GITHUB_TOKEN }}
+        password: ${{ secrets.ACR_SERVICE_PRINCIPAL_PASSWORD }}
         # Container registry server url
-        login-server: https://docker.pkg.github.com/
+        login-server: https://${{ secrets.ACR_REGISTRY }}
     - name: Build and push Docker image
       run: |
         $Now = [DateTime]::Now;
         $ImageTag = "{0}{1:d2}{2:d2}.{3:d2}{4:d2}.{5:d2}{6:d4}" -f $Now.Year, $Now.Month, $Now.Day, $Now.Hour, $Now.Minute, $Now.Second, $Now.Millisecond
         $ImageName = "samples"
-        $RepoName = "microsoft/quantum"
+        $RepoName = "public/quantum/samples"
         
         $LocalTag = "${ImageName}:${ImageTag}"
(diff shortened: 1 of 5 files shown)


cgranade:
So the infrastructure for this works out to be pretty straightforward, and seems to work pretty well at getting Docker images from master up. The trouble is that apparently GitHub Package Registry has an undocumented requirement that users log in to `docker pull` even public packages from public repos (more details at https://github.community/t5/GitHub-Actions/docker-pull-from-public-GitHub-Package-Registry-fail-with-quot/td-p/32782). This means that the new images deploy correctly, but are difficult or impossible to use with remote development, VS Online, or Binder. Since those are our three biggest usecases, I will work to explore other solutions.

+@anpaz-msft

Dockerfile
```diff
@@ -1,46 +1,10 @@
...
-# https://github.com/microsoft/iqsharp/blob/master/images/iqsharp-base/Dockerfile.
-# As per Binder documentation, we choose to use an SHA sum here instead of a
-# tag.
-FROM mcr.microsoft.com/quantum/iqsharp-base:0.10.1912.501
+# This uses the latest Docker image built from the samples repository,
+# defined by the Dockerfile in Build/images/samples.
+FROM docker.pkg.github.com/microsoft/quantum/samples:latest
 
 # Mark that this Dockerfile is used with the samples repository.
 ENV IQSHARP_HOSTING_ENV=SAMPLES_DOCKERFILE
```
anpaz-msft: I wonder if we should have different values for both, maybe this one should be SAMPLES_HOSTED?

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Azure login
+      uses: Azure/login@v1
+      with:
+        creds: ${{ secrets.ACR_CREDENTIALS }}
+    - name: Docker Login
+      run: az acr login -n msquantum
+    - name: Build and push Docker image
+      run: |
+        $Now = [DateTime]::Now;
+        $ImageTag = "{0}{1:d2}{2:d2}.{3:d2}{4:d2}.{5:d2}{6:d4}" -f $Now.Year, $Now.Month, $Now.Day, $Now.Hour, $Now.Minute, $Now.Second, $Now.Millisecond
```
anpaz-msft: would it be better to use the commit id?

Dockerfile
cgranade: Good point, easy enough to change, yeah.

.github/workflows/main.yml
cgranade: Happy to do that as well, it's easily accessible as `${{ github.sha }}`. My only thought was that this gives a strictly increasing ordering. That said, we likely won't need to directly refer to tags other than latest, so let me switch over to SHA sums.

PR #285, opened 2019-12-29
Onboard the H₂ command line sample to the samples browser.

cgranade:


cgranade:
> One minor question: will it be clear to the reader where to look for simulation.sln (since it's several directories up)?

That's a good point, it looks like the `*.sln` files up a folder aren't getting included by the ZIP files on the samples browser. For instance, downloading the Qrng sample code gives me everything I need to run _except_ the sln file:
image

At a minimum, it looks like we need to adapt the instructions. Thanks for catching that, @tcNickolas!

cgranade:
It looks like Visual Studio 2019's "anycode" (Open Folder) mode works, let me modify the instructions accordingly.

PR #292, opened 2020-01-04
Update QML samples, add wine dataset sample

cgranade:
This PR parallels microsoft/quantumlibraries#200 by updating the existing half-moons sample to use the new API features in that PR, and by adding a new sample based on the wine dataset that demonstrates constructing classifier structures in Q#. This PR also prepares both samples for eventual onboarding into the samples browser by adding appropriate README.md files.


cgranade:
Merging this in now, as it's been three weeks without any feedback.

PR #294, opened 2020-01-06
Onboarding bit-flip code to samples browser.

cgranade:
This PR continues the samples browser onboarding process by including the bit-flip code sample into our offerings. That should also help in building on progress in MicrosoftDocs/quantum-docs-pr#652 by allowing us to better link to the sample from documentation.


samples/error-correction/bit-flip-code/BitFlipCode.qs
```diff
@@ -373,5 +374,5 @@ namespace Microsoft.Quantum.Samples.BitFlipCode {
         check(errors[1]);
```
tcNickolas: Would it make sense to use a loop here? Something like `for (error in [X0, X1, X2]) { check(error); }`? Assembling the constants into an array looks weird if that array is never used as an array.

cgranade: Thanks for taking a look, I've modified accordingly.

PR #299, opened 2020-01-08
Add an example of overriding simulator operations

tcNickolas:
Co-Authored-By: Andres Paz <anpaz@microsoft.com>


Samples/runtime/simulator-with-overrides/SimulatorWithOverrides.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <PlatformTarget>x64</PlatformTarget>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Standard" Version="0.10.1911.1607" />
```
cgranade:
```suggestion
    <PackageReference Include="Microsoft.Quantum.Standard" Version="0.10.1912.501" />
```

Samples/runtime/simulator-with-overrides/SimulatorWithOverrides.csproj
```diff
@@ -0,0 +1,12 @@
...
+    <PackageReference Include="Microsoft.Quantum.Development.Kit" Version="0.10.1911.1607" />
```
cgranade:
```suggestion
    <PackageReference Include="Microsoft.Quantum.Development.Kit" Version="0.10.1912.501" />
```

Samples/runtime/simulator-with-overrides/Host.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    class Host
+    {
+        static void Main(string[] args)
+        {
+            using FaultyMeasurementsSimulator qsim = new FaultyMeasurementsSimulator();
+            DoCorrelatedMeasurements.Run(qsim).Wait();
+        }
+    }
+}
```
cgranade:
```suggestion
}

```

Samples/runtime/simulator-with-overrides/Host.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using Microsoft.Quantum.Simulation.Simulators;
+
+namespace Microsoft.Quantum.Samples.SimulatorWithOverrides
+{
+    class Host
+    {
+        static void Main(string[] args)
+        {
+            using FaultyMeasurementsSimulator qsim = new FaultyMeasurementsSimulator();
```
cgranade:
```suggestion
            using var qsim = new FaultyMeasurementsSimulator();
```

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            public M(FaultyMeasurementsSimulator m) : base(m) { }
+
+            /// <summary>
+            /// The actual definition of what the new operation does.
+            /// </summary>
+            public override Func<Qubit, Result> Body {
+                get {
+                    // Get the original M operation to call it and process the results
+                    Func<Qubit, Result> originalMeasurementOperation = base.Body;
```
cgranade:
```suggestion
                    var originalMeasurementOperation = base.Body;
```

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        public class M : QSimM
+        {
+            /// <summary>
+            /// The probability with which the measurement result will be flipped.
+            /// </summary>
+            const double flipProbability = 0.1;
+
+            /// <summary>
+            /// Random numbers generator used to decide when to flip the result.
```
cgranade:
```suggestion
            /// Random number generator used to decide when to flip the result.
```

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+            /// <summary>
+            /// The actual definition of what the new operation does.
+            /// </summary>
+            public override Func<Qubit, Result> Body {
+                get {
+                    // Get the original M operation to call it and process the results
+                    Func<Qubit, Result> originalMeasurementOperation = base.Body;
+
+                    // Get the X gate
+                    gateX = this.Factory.Get<IUnitary<Qubit>>(typeof(X));
```
cgranade: Since this static member is populated on every call to `Body`' accessor, does it need to be a static member, or can it be made into a local variable instead?

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                    // The body of the operation is a lambda
+                    return (qubit =>
+                    {
+                        // Call the original M operation to get correct measurement results
+                        Result originalResult = originalMeasurementOperation(qubit);
```
cgranade:
```suggestion
                        var originalResult = originalMeasurementOperation(qubit);
```

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                        // Flip the measurement result with certain probability
+                        if (rnd.NextDouble() < flipProbability) {
```
cgranade:
```suggestion
                        if (rnd.NextDouble() < flipProbability)
                        {
```

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+                            // Remember to adjust the state of the wave function
+                            gateX.Apply(qubit);
```
cgranade: If I understand correctly, this doesn't actually implement a faulty measurement, but rather a perfect measurement immediately preceded by a symmetric bit-flip channel. For a faulty measurement, you'd likely want something described by the POVM {η 𝟙 / 2 + (1 - η) |0⟩⟨0|, η 𝟙 / 2 + (1 - η) |1⟩⟨1|} where η ≔ `2 * flipProbability`. That POVM represents that with probability 1 - η, you get the correct measurement result, while with probability η, you get a random answer by flipping a coin instead.

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
```diff
@@ -0,0 +1,71 @@
...
+            /// Random numbers generator used to decide when to flip the result.
+            /// </summary>
+            private static readonly System.Random rnd = new System.Random();
+
+            /// <summary>
+            /// The X gate operation used to adjust the state to match flipped measurement results.
+            /// </summary>
+            private static IUnitary<Qubit> gateX;
+
+            public M(FaultyMeasurementsSimulator m) : base(m) { }
```
cgranade: Since you're using `ResetAll` below, you'll likely want to initialize the base `QuantumSimulator` to throw when qubits aren't in the |0⟩ state at release-time. In my experience, faulty measurements cause reset operations to incorrectly reset, such that the simulator will exception out pretty frequently.

Samples/runtime/simulator-with-overrides/Operations.qs
```diff
@@ -0,0 +1,32 @@
...
+                // Measure both qubits; if there is an error introduced during one of the measurements (but not both), the results will diverge
+                if (M(q1) == M(q2)) {
+                    set nSame += 1;
+                }
+
+                // Make sure to return the qubits to 0 state
+                ResetAll([q1, q2]);
+            }
+        }
+        Message($"{nSame} runs out of {nRuns} produced the same results");
```
cgranade:
```suggestion
        Message($"{nSame} runs out of {nRuns} produced the same results.");
```

Samples/runtime/simulator-with-overrides/Operations.qs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// # Summary
+    /// Run a series of experiments that on a perfect state simulator produce matching measurement results.
+    /// If executed on a simulator which introduces errors in measurements, 
+    /// a certain percentage of experiments will produce mismatched results.
+    operation DoCorrelatedMeasurements () : Unit {
+        let nRuns = 100;
+        mutable nSame = 0;
+        for (i in 1 .. nRuns) {
+            using ((q1, q2) = (Qubit(), Qubit())) {
```
cgranade: If you decide to redefine the error model to allow for faulty measurements, then the qubits as allocated at this point in the program may no longer be in the correct state; in that case, you'll prepare one of the other four Bell pairs, rather than the intended (|00⟩ + |11⟩) / √2. As a result, the behavior at line 22 may be different than expected; a comment to that effect may be helpful.

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
tcNickolas: Hmm, I think I did what I wanted to do (have the system collapse to a different state than it wanted to, but report measurement result consistent with what it collapsed to) but described it incorrectly. (The variant you described is one of the things I'd like the competition participants to explore.)

I'll update the descriptions accordingly. Thank you for the clarification!

cgranade: Makes sense, either way is an interesting model after all! Happy to help, in any case.

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
tcNickolas: That's the default behavior of `QuantumSimulator`, I don't think any special initialization is required? (Helped me to find bug in my implementation yesterday, so it's definitely "on")

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
tcNickolas: For this sample I'd really like to keep things as explicit as possible, since people will be using it to start their work with, and figuring out the types of the more unusual variables in a very unfamiliar code base is very tricky.

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
cgranade: Sorry, typo on my part, I should have said "to _not_ throw."

Samples/runtime/simulator-with-overrides/Operations.qs
tcNickolas: I kept the model that keeps measurement results in sync with the state, so this should not be the case. I'm positively looking forward to people experimenting with this kind of behavior on different models :-)

cgranade: Sounds good, hopefully this will spur enough exploration that someone will want to contribute an open systems simulator!

Samples/runtime/simulator-with-overrides/FaultyMeasurementsSimulator.cs
cgranade: I guess my thinking is that the general coding style to use `var` in sample C# code fits well here, too, in that the type of the property is already given at L44. Similarly at L10, the type is already given once with the expression `new FaultyMeasurementSimulator()`, such that using an explicit type is redundant.

That said, I also appreciate the need to make exceptions to general coding guidelines where there's a specific usecase for doing so, such as where there's need to emphasize a particularly unfamiliar type. With respect to L56, for example, I could see expanding the comment and making the local variable type explicit to help readers of this sample understand that the Q# type `Result` maps to a C# type with the same name — that is definitely not obvious at first, given that Q# `Int` maps to C# `long` and that Q# `'T[]` maps to `IQArray<T>`.

PR #309, opened 2020-01-31
Update QML samples to use final API

cgranade:
This PR updates the half-moon and wine samples to use the changes in microsoft/QuantumLibraries#220.


samples/machine-learning/wine/Wine.csproj
```diff
@@ -8,6 +8,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Quantum.MachineLearning" Version="0.10.2001.2808-beta" />
```
anpaz-msft: Does it matter that the SDK doesn't match the nuget package version? I'm thinking the SDK defines what version of Intrinsics and Standard are included (although I could be wrong).

PR #319, opened 2020-02-23
Update qml samples

cgranade:
This PR uses the performance improvements from https://github.com/microsoft/qsharp-runtime/pull/151 and https://github.com/microsoft/QuantumLibraries/pull/234 to add more data to half-moons and to dramatically accelerate the performance of wine.


cgranade:
> So, what are the changes in the data.json file? Is this the output from running the training? Why is it part of the sample?

The data.json file is used to load the training and validation data that the sample acts on. With the improved performance, we can include a larger training set as per @geduardo's suggestion.

PR #320, opened 2020-02-24
Add parallel half-moons sample, update to provide better diagnostic messages.

cgranade:
This PR adds a new half-moons sample that parallelizes over models in the C# host using the `TrainSequentialClassifierAtModel` operation together with Parallel LINQ (PLINQ). This PR also makes a few minor improvements to the existing QML samples, including better console diagnostics.


samples/machine-learning/parallel-half-moons/data.json
```diff
@@ -0,0 +1,774 @@
+{
```
bettinaheim: Is the data the same as for the other sample? If so, does the duplication make sense?

samples/machine-learning/parallel-half-moons/ParallelHalfMoons.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.Quantum.Sdk/0.10.2002.2310-beta">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <PlatformTarget>x64</PlatformTarget>
```
bettinaheim: This is not needed since the Sdk defines this property. 
```suggestion
```

samples/machine-learning/parallel-half-moons/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Prerequisites
+
+- The Microsoft [Quantum Development Kit](https://docs.microsoft.com/quantum/install-guide/).
+
+## Running the Sample
+
+This sample can be run in a number of different ways, depending on your preferred environment.
+
+<!-- TODO
```
bettinaheim: This should probably be removed, since for the parallelized sample going with a C# driver makes more sense.

geduardo:
Sorry, I didn't get to the last PR. I tried both (normal and parallel) and they work fine without errors, parallel being noticeably faster. However, I found the new verbose output a bit overwhelming. It's dificult to read due to the fast changing text in the terminal  and for me isn't very clear what is exactly doing or what the progress is. Tell me if I can help with anything.

cgranade:
> Sorry, I didn't get to the last PR.

No worries at all!

> I tried both (normal and parallel) and they work fine without errors, parallel being noticeably faster.

That's really good to know! I've seen the parallelized version being slower in some cases, especially on systems with small numbers of cores; how many cores did you run on, out of curiousity?

> However, I found the new verbose output a bit overwhelming. It's difficult to read due to the fast changing text in the terminal and for me isn't very clear what is exactly doing or what the progress is. Tell me if I can help with anything.

That's good feedback, thank you! Perhaps it would be good, then, to remove some of the `options::VerboseMessage` calls from the library. That should require any further changes to the samples, though, to implement.

samples/machine-learning/parallel-half-moons/data.json
cgranade: I could remove it from `parallel-half-moons` and change the path used to look for it. The only issue would be with onboarding to the Samples Browser, which expects that all files used by a single sample live in the same directory.

samples/machine-learning/parallel-half-moons/README.md
cgranade: Good point, let me do a pass on comments / README on the new sample in general.

geduardo:
> That's really good to know! I've seen the parallelized version being slower in some cases, especially on systems with small numbers of cores; how many cores did you run on, out of curiousity?
Here you have the specs of my CPU: https://ark.intel.com/content/www/us/en/ark/products/97472/intel-core-i5-7300u-processor-3m-cache-up-to-3-50-ghz.html

I run it with the default settings I think, just using ´dotnet run´.

samples/machine-learning/parallel-half-moons/data.json
bettinaheim: Ah ok, never mind then.

PR #321, opened 2020-02-26
Samples for new QML library.

cgranade:
This PR deploys all three samples for the QML library to master.


samples/machine-learning/parallel-half-moons/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Open the folder containing this sample in Visual Studio ("Open a local folder" from the Getting Started screen or "File → Open → Folder..." from the menu bar) and set `HalfMoons.csproj` as the startup project.
+Press Start in Visual Studio to run the sample.
+
+## Manifest
+
+- [Training.qs](https://github.com/microsoft/Quantum/blob/master/samples/machine-learning/half-moons/Training.qs): Q# code implementing quantum operations for this sample.
+- [Host.cs](https://github.com/microsoft/Quantum/blob/master/samples/machine-learning/half-moons/Host.cs): C# code to load data, and to interact with and print out results of the Q# operations for this sample.
+- [ParallelHalfMoons.csproj](https://github.com/microsoft/Quantum/blob/master/samples/machine-learning/half-moons/HalfMoons.csproj): Main C# project for the sample.
+- [data.json](https://github.com/microsoft/Quantum/blob/master/samples/machine-learning/half-moons/data.json): JSON-serialized training and validation data used by this sample.
```
geduardo: Are this URLs right? I think they should point to the parallel-half-moon folder instead

samples/machine-learning/parallel-half-moons/Host.cs
```diff
@@ -0,0 +1,157 @@
...
+                .WithExecutionMode(ParallelExecutionMode.ForceParallelism)
+                // Many of the same LINQ methods are defined for PLINQ queries
+                // as well as IEnumerable objects, so we can go on and run
+                // the training loop in parallel by selecting on the start point.
+                .Select(
+                    (startPoint, idxStartPoint) =>
+                    {
+                        // Since we want each start point to run on its own
+                        // instance of the full-state simulator, we create a new
+                        // instance here, using C# 8's "using var" syntax to
```
alan-geller:
```suggestion
                        // instance here, using C# 8's "using" syntax to
```
You can actually use `using` this way with an explicit type, not just `var`.

samples/machine-learning/parallel-half-moons/README.md
cgranade: Good catch, thank you! Since this is from a protected branch, I'll complete this PR for now, then will update that in the next PR.

samples/machine-learning/parallel-half-moons/Host.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            // Convert samples to Q# form.
+            var samples = new QArray<QArray<double>>(data.TrainingData.Features.Select(vector => new QArray<double>(vector)));
+
+            // Once we have the data loaded and have initialized our target machine,
+            // we can then use that target machine to train a QCC classifier.
+            var (optimizedParameters, optimizedBias, nMisses) = parameterStartingPoints
+                // We can use parallel LINQ (PLINQ) to convert the IEnumerable
+                // over starting points into a parallelized query.
+                .AsParallel()
```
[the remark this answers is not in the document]
cgranade: Absolutely, and we've seen that in practice; this sample was suggested in part by @alexeib2 to show how to use additional cores in cases where there are enough cores available to gain from parallelization.

samples/machine-learning/parallel-half-moons/Host.cs
cgranade: Good point, thank you! As with @geduardo's feedback, I'll defer to the next PR so that I can close this protected branch.

PR #325, opened 2020-03-09
Remove qdk-migrate.ps1

anpaz-msft:
This file was tagged by our  SDL tools because the signature includes an offensive word in english.
This script was to help migrate people to 0.3.  I think it's not needed anymore.


cgranade:
Where's a facepalm GIF when you need it? Thanks for taking care of this, I agree, it's no longer needed. Approving accordingly.

anpaz-msft:
I know, right?

PR #327, opened 2020-03-13
Add link to iterative phase estimation docs to sample readme

tcNickolas:
For a bit of context, I was searching for "iterative phase estimation" and this sample was both the first result (docs view) and the third one (GitHub view), but it doesn't point to the relevant page in our docs. It seems like including such link would be a good idea.


tcNickolas:
Shall we merge the PR? CI is stuck but, seeing how this is readme-only change, I don't think it's going to break anything.

PR #335, opened 2020-04-10
Update to 0.11.2003.3107.

cgranade:
This PR updates all samples to 0.11.2003.3107, as well as the Dockerfile used to generate the samples image.


samples/algorithms/database-search/Database Search.ipynb
```diff
@@ -53,9 +53,9 @@
...
       "text/html": [
-       "<h4><i class=\"fa fas fa-terminal\"></i> Microsoft.Quantum.Canon.ControlledOnInt <a href=\"/home/jovyan/.nuget/packages/microsoft.quantum.standard/0.9.1908.2906/lib/netstandard2.1/Microsoft.Quantum.Standard.dll\"><i class=\"fa fas fa-code\"></i></a></h4><h1>Summary</h1>\n",
+       "<h4><i class=\"fa fas fa-terminal\"></i> Microsoft.Quantum.Canon.ControlledOnInt <a href=\"C:/Users/Chris/.nuget/packages/microsoft.quantum.standard/0.11.2003.3107/lib/netstandard2.1/Microsoft.Quantum.Standard.dll\"><i class=\"fa fas fa-code\"></i></a></h4><h1>Summary</h1>\n",
```
vxfield, line 58: Shouldn't we clean the outputs cells in this case?

samples/arithmetic/AdderExample.ipynb
```diff
@@ -361,16 +361,16 @@
    "outputs": [
     {
      "data": {
-      "application/json": "{\"rows\":[{\"@type\":\"@tuple\",\"Item1\":\"iqsharp\",\"Item2\":\"0.10.2001.2831\"},{\"@type\":\"@tuple\",\"Item1\":\"Jupyter Core\",\"Item2\":\"1.2.29939.0\"},{\"@type\":\"@tuple\",\"Item1\":\".NET Runtime\",\"Item2\":\".NETCoreApp,Version=v3.0\"}]}",
+      "application/json": "{\"rows\":[{\"@type\":\"@tuple\",\"Item1\":\"iqsharp\",\"Item2\":\"0.11.2003.3107\"},{\"@type\":\"@tuple\",\"Item1\":\"Jupyter Core\",\"Item2\":\"1.2.36563.0\"},{\"@type\":\"@tuple\",\"Item1\":\".NET Runtime\",\"Item2\":\".NETCoreApp,Version=v3.1\"}]}",
       "text/html": [
-       "<table><thead><tr><th>Component</th><th>Version</th></tr></thead><tbody><tr><td>iqsharp</td><td>0.10.2001.2831</td></tr><tr><td>Jupyter Core</td><td>1.2.29939.0</td></tr><tr><td>.NET Runtime</td><td>.NETCoreApp,Version=v3.0</td></tr></tbody></table>"
+       "<table><thead><tr><th>Component</th><th>Version</th></tr></thead><tbody><tr><td>iqsharp</td><td>0.11.2003.3107</td></tr><tr><td>Jupyter Core</td><td>1.2.36563.0</td></tr><tr><td>.NET Runtime</td><td>.NETCoreApp,Version=v3.1</td></tr></tbody></table>"
```
vxfield, line 366: Same here, checking if we should clean the output cells...

samples/getting-started/intro-to-iqsharp/Notebook.ipynb
```diff
@@ -325,9 +325,9 @@
    "outputs": [
     {
      "data": {
-      "application/json": "1",
+      "application/json": "0",
```
vxfield, line 328: Same here, checking if we should clean the output cells from the commit

cgranade:
> Made a few comments on whether we want to restrict the commit only to the changes to the versions and necessary fixes and leave the output cells blank.

We've generally not left the output cells blank so that they can be viewed using GitHub's integrated notebook viewer, or with https://nbviewer.jupyter.org/. For the katas, the policy tends to be different given the different goals (there's not a lot of point in reading a kata as a static document, after all).

samples/arithmetic/AdderExample.ipynb
cgranade: This one is especially good to keep, as it lets readers of a notebook know what version was used when it was last run. It's a best-practice that the QuTiP team has advocated for, for example.

samples/algorithms/database-search/Database Search.ipynb
cgranade: Agreed here, we shouldn't have per-user paths in the output. I ran locally this time, rather than in the devcontainer since the container still needed updated; hence the change in paths. I'll fix that back to `jovyan`. Thanks!

PR #343, opened 2020-04-30
Change standalone Qrng to use RUS loop.

cgranade:


samples/getting-started/qrng/Qrng.qs
```diff
@@ -16,18 +16,25 @@ namespace Qrng {
     }
     
     @EntryPoint()
-    operation Main() : Int {
+    operation RunMain() : Int {
```
rmshaffer, line 19: Since you're changing this anyway - what about `GenerateRandomNumber`?

samples/getting-started/qrng/Qrng.qs
```diff
@@ -16,18 +16,25 @@ namespace Qrng {
...
         let max = 10;
```
geduardo, line 19: What about max = 50? For uniformity with the host codes and also it gets the probability of failing a bit lower (~20% instead of 33%)

samples/getting-started/qrng/Qrng.qs
cgranade: Sounds good, will do!

geduardo: Mmm, I like `Main()` here due to the nature of the operation, being executed by running the whole file. It resembles the conventions used in other languages.

cgranade: That's a good point, @geduardo. That said, in general, the Q# style guide suggests nouns should be reserved for functions and UDTs. While the style guide was drafted before the standalone feature, I'm inclined keep to that here as well, to go with @rmshaffer's suggestion of a more descriptive name. I think that while `Main` would be reflective of general-purpose languages, this reflects better that Q# programs act in a more accelerator-like model. Happy for further discussion on this, though!

samples/getting-started/qrng/Qrng.qs
cgranade: Good point, let me go on and make the standalone version more uniform with the interoperability version.

PR #359, opened 2020-05-20
Trivial changes to trigger builder

geduardo:
I did some trivial changes to trigger the builder. I also noticed that many samples don't have any YAML file, is this on purpose or should we add them?


cgranade:
> I did some trivial changes to trigger the builder. I also noticed that many samples don't have any YAML file, is this on purpose or should we add them?

We haven't yet onboarded all samples to have the right metadata; that's ongoing work as we refactor each sample to have the right format.

PR #381, opened 2020-07-06
Add RUS sample (#12394)

guenp:
Initial new version of RUS sample that breaks up circuit into two parts (red and blue, see image) and only runs second (blue) part if ancilla measurement in the first part (red) returns `Zero`.
Draft version todo's
[x] docs
[x] unit tests
[o] add hardware-compatible sample -> moving this to follow up deliverable (TBD)


samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
+namespace Microsoft.Quantum.Samples.RepeatUntilSuccess {
```
cgranade, line 1: We should make sure this has an appropriate copyright header; should be able to use one from similar files elsewhere in the samples.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Preparation;
+
+    /// Example of a Repeat-until-success circuit
```
cgranade:
```suggestion
    /// # Summary
    /// Example of a Repeat-until-success circuit.
    ///
    /// # Description
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+    /// The circuit is executed on a "target" qubit using an "ancilla" and
```
cgranade, line 15: In general, I'd suggest avoiding calling Q# programs (especially those with measurement adaptivity) circuits, since that can confuse with lower-level non-adaptive constructs.

cgranade, line 15: In general, Q# style has been to use "auxiliary" in strong preference to "ancillary" given the racial and misogynistic implications of the latter.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+    /// If this fails, rerun the circuit up to <limit> times.
+    @EntryPoint()
+    operation ApplyRzArcTan2(
+        inputValue : Bool,
+        inputBasis: Pauli,
+        limit: Int
+    ) : (Bool, Result, Int) {
+        using ((ancilla, resource, target) = (Qubit(), Qubit(), Qubit())) {
+            // Prepare qubits
+            PrepareXZero(ancilla);
```
cgranade, line 29: The API documentation comment for `PrepareXZero` indicates that it expects its input to be in the |1⟩ state, but here, `ancilla` starts off in |0⟩ as it is freshly allocated. In general, the Q# verb "Prepare" indicates that an operation expects a qubit input to be in the |0⟩ state, so this may be an issue with the name and documentation comments for `PrepareXZero`.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                } else { // |1X>, skip Part 2
+                    // Reset ancilla from |1> to |+>
+                    PrepareXZero(ancilla);
+                }
+                set done = (result2 == Zero or numIter >= limit); //|00>
+                set numIter = numIter + 1;
+            }
+            until (done)
+            fixup {}
```
cgranade, line 60: The fixup block at the end is now optional, such that I'd recommend leaving it off when it's empty.
```suggestion
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+        limit: Int
+    ) : (Bool, Result, Int) {
+        using ((ancilla, resource, target) = (Qubit(), Qubit(), Qubit())) {
+            // Prepare qubits
+            PrepareXZero(ancilla);
+            PrepareXZero(resource);
+            PrepareValueForBasis(inputValue, inputBasis, target);
+
+            // Initialize results to One by default.
+            mutable result1 = One;
```
cgranade, line 34: It looks like this declaration can be brought inside the loop (replacing `set` at line 35), and made immutable.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+            mutable result2 = One;
```
cgranade, line 35: Similarly here, you can avoid making a mutable declaration by bringing `result2` inside the loop.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+                    // Reset ancilla from |1> to |+>
+                    PrepareXZero(ancilla);
+                }
+                set done = (result2 == Zero or numIter >= limit); //|00>
+                set numIter = numIter + 1;
+            }
+            until (done)
+            fixup {}
+
+            let success = (result1 == Zero and result2 == Zero);
```
cgranade, line 62: Ah, I see; this is why `result1` and `result2` are defined outside the loop...

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+
+            // Rz(2.0*ArcTan(2.0), target);
+            let result = Measure([inputBasis], [target]);
+
+            Reset(target);
+            Reset(resource);
+            Reset(ancilla);
+
```
cgranade, line 70: These may not be needed as of 0.12, as releasing measured qubits is now allowed.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+                }
+                set done = (result2 == Zero or numIter >= limit); //|00>
+                set numIter = numIter + 1;
+            }
+            until (done)
+            fixup {}
+
+            let success = (result1 == Zero and result2 == Zero);
+
+            // Rz(2.0*ArcTan(2.0), target);
```
cgranade, line 64: What does this comment indicate?

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+
+            Reset(target);
+            Reset(resource);
+            Reset(ancilla);
+
+            return ( success, result, numIter );
+        }
+    }
+
+    // Prepare qubit in either |0> or |1> for the given basis
```
cgranade, line 75: I'd suggest making thiss comment into an API documentation comment (`///`) so that it is available to intelliSense elsewhere in the file.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,110 @@
...
+    operation PrepareValueForBasis(
+        inputValue : Bool,
+        inputBasis : Pauli,
+        input: Qubit) : Unit {
```
cgranade, line 79:
```suggestion
        input : Qubit
    )
    : Unit {
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: will do!

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: I'll try to be more explicit about which part is the program and which part is the circuit.

guenp: Fixed!

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: What naming would you suggest in this case?
Would `SetXZeroFromOne` work?

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: I'll make success mutable instead.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: Hmm good point, it still fails for me locally if I don't include this so I'd prefer to keep it in in case there are other users who are not on the bleeding edge

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: This was part of the old sample. I'm not sure if I should leave it in or not, I'll need to ask @apaetz

swernli:
This seems like an excellent sample for demonstrating the algorithm on Q#, but I don't think this would work as written when targeting partner execution due to some translation limitations. Is the goal here to produce a sample that works across both simulation and execution or is this more focused on a sample of clean, efficient Q#? I just want to make sure I understand the goal before I get too deep into feedback along the wrong axis :).

samples/algorithms/repeat-until-success/RepeatUntilSuccess.csproj
```diff
@@ -0,0 +1,8 @@
+﻿<Project Sdk="Microsoft.Quantum.Sdk/0.11.2004.2825">
```
swernli, line 1: If you update the version here to the latest release, `0.12.20070124` then you would get the newer behavior Chris referred to, such as the allowance to release qubits that have been measured but not explicitly reset.

guenp:
> This seems like an excellent sample for demonstrating the algorithm on Q#, but I don't think this would work as written when targeting partner execution due to some translation limitations. Is the goal here to produce a sample that works across both simulation and execution or is this more focused on a sample of clean, efficient Q#? I just want to make sure I understand the goal before I get too deep into feedback along the wrong axis :).

@swernli Is there documentation anywhere on these translation limitations? I suggest we could make two versions.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
swernli: I did not know this, thanks for raising awareness Chris! I found this helpful explanation online as well: https://grammarist.com/usage/ancillary-vs-auxiliary/

swernli:
We don't have the documentation yet, though there is work ongoing to capture the supported features as quantum processor capabilities (see https://github.com/microsoft/qsharp-compiler/pull/488), which would allow developers to get compile time feedback on whether their code would work as-written for a particular target. As for two samples vs one, I'll let @cgranade weigh in there. I think the samples included here have thus far been primarily for pure Q#, but I don't know if there is any expectation around shifting that focus to also include hardware compatible samples.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
cgranade: The project file should set the version number, such that users don't need to do anything locally to upgrade to 0.12. From that perspective, I don't think requiring the latest Quantum Development Kit version is necessarily the bleeding edge; the case is a bit different for IQ#, though, as that requires global installation rather than project-local installation.

guenp: Ok that works for me

guenp:
> We don't have the documentation yet, though there is work ongoing to capture the supported features as quantum processor capabilities (see [microsoft/qsharp-compiler#488](https://github.com/microsoft/qsharp-compiler/pull/488)), which would allow developers to get compile time feedback on whether their code would work as-written for a particular target. As for two samples vs one, I'll let @cgranade weigh in there. I think the samples included here have thus far been primarily for pure Q#, but I don't know if there is any expectation around shifting that focus to also include hardware compatible samples.

@cgranade I think it would be useful to have hardware-compatible examples in the samples for those running stuff on Azure quantum. My suggestion would be to make two versions and have an input argument `--hw-compatible` or so. Any thoughts?

cgranade:
> We don't have the documentation yet, though there is work ongoing to capture the supported features as quantum processor capabilities (see [microsoft/qsharp-compiler#488](https://github.com/microsoft/qsharp-compiler/pull/488)), which would allow developers to get compile time feedback on whether their code would work as-written for a particular target. As for two samples vs one, I'll let @cgranade weigh in there. I think the samples included here have thus far been primarily for pure Q#, but I don't know if there is any expectation around shifting that focus to also include hardware compatible samples.

I think a key principle needs to be that adapting samples to be hardware-compatible shouldn't make it harder to understand the samples or what they teach about Q# and the Quantum Development Kit. Thus, I'm all for adapting existing samples to be compatible with current capabilities if we can do so in a fairly nice way, otherwise I would suggest splitting into two samples.

We have some precedent for this, in fact, with the host-language interoperability samples. The Python interoperability feature has some limitations that make it hard to use for all existing samples, such that there's a dedicated samples section under `interoperability/python`. 

> @cgranade I think it would be useful to have hardware-compatible examples in the samples for those running stuff on Azure quantum. My suggestion would be to make two versions and have an input argument `--hw-compatible` or so. Any thoughts?

Agreed with making two versions, as per the above. I don't know that we could make them switch with an input argument, though, given that Q# applications compile as whole programs; @swernli could likely clarify here.

swernli:
@cgranade That all sounds reasonable to me.

@guenp Given that we plan on having a separate sample for hardware targets, I don't think anything here needs to be adjusted for translation compatibility.

I can tell it's a great sample, because I've already learned a few new tricks from it! Well done!

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,116 @@
...
+    /// Rz(arctan(2)) gate (also known as V_3 gate) on the target qubit.
+    /// If this fails, the program reruns the circuit up to <limit> times.
+    @EntryPoint()
+    operation ApplyRzArcTan2(
+        inputValue : Bool,
+        inputBasis: Pauli,
+        limit: Int
+    ) : (Bool, Result, Int) {
+        using ((auxiliary, resource, target) = (Qubit(), Qubit(), Qubit())) {
+            /// Prepare auxiliary and resource qubits in |+> state
```
msoeken, line 33:
```suggestion
            // Prepare auxiliary and resource qubits in |+〉state
```

I am suggesting to use `//` comments for simple line comments and use `///` for potentially parsable documentation comments.

I am also suggesting the use of unicode in the sample comments to get closer to mathematical typesetting.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,116 @@
...
+                    // Reset auxiliary from |1> to |+>
+                    SetXZeroFromOne(auxiliary);
+                }
+                set done = (success or numIter >= limit);
+                set numIter = numIter + 1;
+            }
+            until (done);
+
+            let result = Measure([inputBasis], [target]);
+            return ( success, result, numIter );
```
msoeken, line 69:
```suggestion
            return (success, result, numIter);
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,116 @@
...
+        ) : Unit {
+            if (inputValue) {
+                X(input);
+            }
+            PrepareQubit(inputBasis, input);
+    }
+
+    /// Prepare qubit in |+> state given it is in the |1> state
+    operation SetXZeroFromOne(target : Qubit) : Unit {
+        X(target); // Flip to |0>
```
msoeken, line 87:
```suggestion
        AssertQubit(One, target);
        X(target); // Flip to |0〉
```

I suggest to add a precondition that checks whether `target` is in expected state |1〉

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,118 @@
...
+    operation ApplyRzArcTan2(
+        inputValue : Bool,
+        inputBasis: Pauli,
+        limit: Int
+    ) : (Bool, Result, Int) {
+        using ((auxiliary, resource, target) = (Qubit(), Qubit(), Qubit())) {
+            // Prepare auxiliary and resource qubits in |+〉state
+            SetXZeroFromOne(auxiliary);
+            SetXZeroFromOne(resource);
+            /// Prepare target qubit in |0> or |1> state, depending on input value
```
msoeken, line 37:
```suggestion
            // Prepare target qubit in |0〉or |1〉 state, depending on input value
```
There are some more occurrences of `///` comments and places where unicode can be useful in the file, just giving one as an example.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,118 @@
...
+            PrepareValueForBasis(inputValue, inputBasis, target);
+
+            /// Initialize results to One by default.
```
msoeken, line 40:
```suggestion
            // Initialize results to One by default.
```

samples/algorithms/repeat-until-success/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Manifest
+
+- **repeat-until-success/**
+  - [RepeatUntilSuccess.csproj](./RepeatUntilSuccess.csproj): Main C# project for the example.
+  - [Program.cs](./Program.cs): C# code to call the operations defined in Q# and perform classical post-processing.
+  - [RepeatUntilSuccess.qs](./RepeatUntilSuccess.qs): The Q# implementation of the RUS algorithm.
+
+## Example run ##
```
msoeken, line 37:
```suggestion
## Example run
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: I believe @cgranade actually asked me to put these in, perhaps we should discuss what we want to do before I change everything back.

cgranade: Apologies for the lack of clarity on my part. In any case, `//` comments should be used for prose as @msoeken says, while `///` comments should be used for API documentation. The latter get picked up by documentation generation tools and by intelliSense when they immediately precede a namespace, operation, function, or user-defined type declaration, but have no special meaning elsewhere. For that reason, I suggest using `//` comments  rather than `///` comments in contexts where documentation comments aren't expected by the compiler (e.g.: within a callable body, between callable declarations, and so forth).

Both are useful in a sample, and serve different purposes, in that `//` comments help establish the prose for a sample, while `///` comments provide contextual reference information.

guenp: Thanks, that makes sense. Posting these docs here for reference: https://docs.microsoft.com/en-us/quantum/user-guide/using-qsharp/file-structure#documentation-comments

guenp:
> @cgranade That all sounds reasonable to me.
> 
> @guenp Given that we plan on having a separate sample for hardware targets, I don't think anything here needs to be adjusted for translation compatibility.
> 
> I can tell it's a great sample, because I've already learned a few new tricks from it! Well done!

Thanks @swernli! We discussed during scrum and we decided to wait with implementing a hardware-compatible sample in our public repo until we go live with the Azure Quantum public preview. I'll follow up with you on guidance for this when that deliverable gets closer :)

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+                    Z(resource); // Reset resource from |-⟩ to |+⟩
+                    Adjoint Z(target); // Correct effective Z rotation on target
+                }
+            } else { //1X
+                // Set auxiliary and resource qubit back to |+⟩
+                Z(auxiliary);
+                Reset(resource);
+                H(resource);
+            }
+            set done = (success or numIter >= limit);
```
msoeken, line 83:
```suggestion
            set done = success or (numIter >= limit);
```
I find it easier to read with the parens set like this.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+            set numIter = numIter + 1;
+        }
+        until (done);
+        return (success, numIter);
+    }
+
+    /// Initialize axiliary and resource qubits in |+⟩, target in |0⟩ or |1⟩
+    operation InitializeQubits(
```
cgranade, line 91: In general, Q# style guide suggests putting classical inputs (Pauli and Bool, in this case) before quantum inputs.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+    ) : (Bool, Result, Int) {
+        using ((auxiliary, resource, target) = (Qubit(), Qubit(), Qubit())) {
+            // Initialize qubits to starting values (|+⟩, |+⟩, |0⟩/|1⟩)
+            InitializeQubits(auxiliary, resource, target, inputBasis, inputValue);
+            let (success, numIter) = ApplyRzArcTan2(auxiliary, resource, target, inputBasis, inputValue, limit);
+            let result = Measure([inputBasis], [target]);
+            // From version 0.12 it is no longer necessary to release qubits in zero state.
+            Reset(target);
+            Reset(resource);
+            Reset(auxiliary);
```
cgranade, line 40:
```suggestion
            ResetAll([target, resource, auxiliary]);
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+            // Assert valid starting states for all qubits
+            AssertMeasurement([PauliX], [auxiliary], Zero, "Auxiliary qubit is not in |+⟩ state.");
+            AssertMeasurement([PauliX], [resource], Zero, "Resource qubit is not in |+⟩ state.");
+            AssertQubitIsInState(target, inputBasis, inputValue);
+
+            // Run Part 1 of the program.
+            let result1 = ApplyAndMeasurePart1(auxiliary, resource);
+            // We'll only run Part 2 if Part 1 returns Zero.
+            // Otherwise, we'll skip and rerun Part 1 again.
+            if (result1 == Zero) { //0X
```
cgranade, line 69: Does "X" mean a wildcard here?

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+        H(resource);
+
+        // Prepare target qubit in |0⟩ or |1⟩ state, depending on input value
+        if (inputValue) {
+            X(target);
+        }
+        PrepareQubit(inputBasis, target);
+    }
+
+    /// Part 1 of RUS circuit (red)
```
cgranade, line 109: Would it be possible to expand this a little bit to use the Markdown format for API documentation comments?

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+    operation ApplyAndMeasurePart1(
+        auxiliary: Qubit,
+        resource: Qubit
+    ) : Result {
+        T(auxiliary);
+        CNOT(resource, auxiliary);
+        Adjoint T(auxiliary);
```
cgranade, line 116:
```suggestion
        within {
            T(auxiliary);
        } apply {
            CNOT(resource, auxiliary);
        }
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+
+        return Measure([PauliX], [auxiliary]);
+    }
+
+    /// Part 2 of RUS circuit (blue)
```
cgranade, line 121: Similarly here, good to include a full API documentation comment if possible.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+    /// The program executes a circuit on a "target" qubit using an "auxiliary" and 
+    /// "resource" qubit. The circuit consists of two parts (red and blue in image).
+    /// The goal is to measure Zero for both the auxiliary and resource qubit.
+    /// If this succeeds, the program will have effectively applied an 
+    /// Rz(arctan(2)) gate (also known as V_3 gate) on the target qubit.
+    /// If this fails, the program reruns the circuit up to <limit> times.
+    @EntryPoint()
+    operation CreateQubitsAndApplyRzArcTan2(
+        inputValue : Bool,
+        inputBasis: Pauli,
```
cgranade, line 29: Very minor, but I'd suggest in general placing a space between identifier names and their types.
```suggestion
        inputBasis : Pauli,
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+    operation AssertQubitIsInState(
+        target: Qubit,
+        inputBasis: Pauli,
+        inputValue: Bool
+    ) : Unit {
+        if (inputValue) {
+            AssertMeasurement([inputBasis], [target], One, "Qubit is not in 1 state for given input basis.");
+        } else {
+            AssertMeasurement([inputBasis], [target], Zero, "Qubit is not in 0 state for given input basis.");
+        }
```
cgranade, line 144:
```suggestion
        AssertMeasurement(
            [inputBasis], [target], inputValue ? One | Zero,
            $"Qubit is not in {inputValue ? One | Zero} state for given input basis."
        );
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// # Summary
+    /// Example of a Repeat-until-success algorithm implementing a circuit 
+    /// that achieves exp(i⋅ArcTan(2)⋅Z) by Paetznick & Svore. 
+    /// Gate exp(i⋅ArcTan(2)⋅Z) is also know as V gate.
+    /// # References
+    /// - [ *Adam Paetznick, Krysta M. Svore*,
+    ///     Quantum Information & Computation 14(15 & 16): 1277-1301 (2014)
+    ///   ](https://arxiv.org/abs/1311.1074)
+    /// For circuit diagram, see file RUS.png (to be added to README).
```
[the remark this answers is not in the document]
guenp: Yes, thanks for noticing!

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    /// The program executes a circuit on a "target" qubit using an "auxiliary" and 
+    /// "resource" qubit. The circuit consists of two parts (red and blue in image).
+    /// The goal is to measure Zero for both the auxiliary and resource qubit.
+    /// If this succeeds, the program will have effectively applied an 
+    /// Rz(arctan(2)) gate (also known as V_3 gate) on the target qubit.
+    /// If this fails, the program reruns the circuit up to <limit> times.
+    @EntryPoint()
+    operation CreateQubitsAndApplyRzArcTan2(
+        inputValue : Bool,
```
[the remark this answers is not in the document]
guenp: Great tip, thanks!

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: Yes, I'll replace with `+` to make it more clear.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,222 @@
...
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Preparation;
+    open Microsoft.Quantum.Diagnostics;
+
+    /// # Summary
+    /// Example of a Repeat-until-success algorithm implementing a circuit 
+    /// that achieves exp(i⋅ArcTan(2)⋅Z) by Paetznick & Svore. 
+    /// Gate exp(i⋅ArcTan(2)⋅Z) is also know as V gate.
```
cgranade, line 13:
```suggestion
    /// The exp(𝑖 ArcTan(2) 𝑍) operation is also known as the "𝑉 gate."
```

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,222 @@
...
+    /// # References
+    /// - [ *Adam Paetznick, Krysta M. Svore*,
+    ///     Quantum Information & Computation 14(15 & 16): 1277-1301 (2014)
+    ///   ](https://arxiv.org/abs/1311.1074)
+    /// For circuit diagram, see file RUS.png.
+    ///
+    /// # Input
+    /// ## inputBasis
+    /// Pauli basis to prepare input qubit in
```
cgranade, line 22: It looks like this is incomplete?

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
```diff
@@ -0,0 +1,222 @@
...
+    /// ## limit
+    /// Integer limit to number of repeats of circuit
+    /// ## auxiliary
+    /// Auxiliary qubit
+    /// ## resource
+    /// Resource qubit
+    /// ## target
+    /// Target qubit
+    operation InitializeQubits(
+        inputBasis: Pauli,
```
cgranade, line 142: [nit] Elsewhere you've used spaces around `:`s, it may be good to do so here as well.

samples/algorithms/repeat-until-success/RepeatUntilSuccess.qs
guenp: I don't think so, but the grammar is a bit awkward so I can fix that.

PR #385, opened 2020-07-12
Bumping electron version because of (unlikely) security issue according to Dependabot 

RolfHuisman:
Dependabot wanted to bump the electron version because of a possible security issue.
While its true, its unlikely the package like this would be used in production (since its an explanation sample), its a good practice to update. Bumping electron however broke two samples because of changed defaults. Hunted that one down.


samples/chemistry/LithiumHydrideGUI/main.js
```diff
@@ -18,7 +18,10 @@ function createWindow () {
   // Create the browser window.
   mainWindow = new BrowserWindow({
       width: 800, height: 600,
-      autoHideMenuBar: true
+      autoHideMenuBar: true,
+      webPreferences: {
```
cgranade, line 22: Thanks for finding this one, @RolfHuisman...!

PR #399, opened 2020-08-10
Guenp/18375 syndrome

guenp:
Add PseudoSyndrome sample based on @apaetz's version in the private repo. Modified to use a Python script as application layer and CLI script instead of C#.

This is still a Draft PR - one way to enrich this sample would be to introduce random errors and show how this affects the auxiliary qubit. @cgranade @rmshaffer @msoeken  I'm curious to hear your thoughts on this, based on how we know these samples are being used and by whom, it may or may not be useful.


samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Samples.Hardware.Syndrome
+{
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Preparation;
+    open Microsoft.Quantum.Measurement;
+	open Microsoft.Quantum.Arrays;
```
guenp, line 10: Oops, something went wrong with indentation here, will fix.

samples/error-correction/syndrome/Syndrome.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.Quantum.Sdk/0.12.20070124">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
```
rmshaffer, line 6: Since you have no operation in your `.qs` file marked as `@EntryPoint()`, this should probably be a library project rather than an executable project.
```suggestion
  <PropertyGroup>
    <TargetFramework>netstandard2.1</TargetFramework>
  </PropertyGroup>
```
I'm assuming you only added this file so that you get IntelliSense, etc.? As a complete aside, once https://github.com/microsoft/qsharp-compiler/pull/560 is done, you'll be able to get language server functionality in VS/VS Code without creating a `.csproj` at all.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+            for ((qubit, value, basis) in Zip3(block, input_values, encoding_bases)) {
+                Prepare(qubit, value, basis);
+            }
+            H(auxiliary);
+            for (index in indexes) {
+                ControlledPauli(encoding_bases[index], auxiliary, block[index]);
+            }
+            let auxiliary_result = Measure([PauliX], [auxiliary]);
+            let data_result = ForEach(BasisMeasure, Zip(encoding_bases, block));
+            Reset(auxiliary);
```
rmshaffer, line 90: Could this just be `MResetX(auxiliary)` instead of separate measurement and reset?
```suggestion
            let auxiliary_result = MResetX(auxiliary);
            let data_result = ForEach(BasisMeasure, Zip(encoding_bases, block));
```

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// ## input_values
+    /// List of boolean values for data qubits
+    /// ## encoding_bases
+    /// List of Pauli bases
+    /// ## indexes
+    /// List of indexes on which to apply controlled pauli operators
+    operation SamplePseudoSyndrome (
+            input_values: Bool[],
+            encoding_bases: Pauli[], 
+            indexes: Int[]
```
rmshaffer, line 78: Since these arrays all must be the same length (if I'm understanding correctly), would it make sense instead for this to take an array of tuples like `(Bool, Pauli, Int)[]`?

samples/error-correction/syndrome/syndrome.py
```diff
@@ -0,0 +1,38 @@
...
+    prog="PseudoSyndrome",
+    description="Program for running a PseudoSyndrome circuit that detects random errors on qubits in gate sequence."
+)
+parser.add_argument("-q", "--qubits", type=int, default=1)
+
+if __name__ == "__main__":
+    args = parser.parse_args()
+    paulis = ["PauliX", "PauliY", "PauliZ"]
+    indexes = list(range(args.qubits))
+    np.random.shuffle(indexes)
```
rmshaffer, line 27: What is the reason to shuffle `indexes` if you are already choosing each of the `input_values` and `encoding_bases` at random?

samples/error-correction/syndrome/syndrome.py
```diff
@@ -0,0 +1,38 @@
...
+
+parser = argparse.ArgumentParser(
+    prog="PseudoSyndrome",
+    description="Program for running a PseudoSyndrome circuit that detects random errors on qubits in gate sequence."
+)
+parser.add_argument("-q", "--qubits", type=int, default=1)
+
+if __name__ == "__main__":
+    args = parser.parse_args()
+    paulis = ["PauliX", "PauliY", "PauliZ"]
```
rmshaffer, line 25: Could you use `[qsharp.Pauli.X, qsharp.Pauli.Y, qsharp.Pauli.Z]` here?

samples/error-correction/syndrome/Syndrome.csproj
guenp: Great point. I added this because I was initially planning to have an `Entrypoint`, but then decided against it. Indeed, I could just leave it out. Thanks for the suggestion.

samples/error-correction/syndrome/Syndrome.qs
guenp: I personally think it creates a clearer interface to separate them out, but I see your point. How about add an assertion to check that those three lists are indeed the same length?

samples/error-correction/syndrome/Syndrome.qs
guenp: I would have to double check this but I believe the auxiliary qubit and data qubits are entangled so resetting the auxiliary qubit before measuring the data qubits may have unwanted effects on the data qubits. However we already measured the auxiliary qubit at this point so it might not matter.

samples/error-correction/syndrome/syndrome.py
guenp: Yes! Thanks!

samples/error-correction/syndrome/syndrome.py
guenp: It shuffles the order in which the controlled Paulis are applied. In this case I agree that it doesn't really matter indeed but I thought this would be relevant in for when I have a set of fixed couplings between qubits in IRL hardware. @apaetz thoughts?

samples/error-correction/syndrome/Syndrome.qs
rmshaffer: Makes sense to me!

samples/error-correction/syndrome/Syndrome.qs
rmshaffer: Got it. Yes I would expect that the measurement destroys the entanglement, but it’s worth double-checking.

samples/error-correction/syndrome/syndrome.py
rmshaffer: Oh, thanks, I somehow missed that! In that case, perhaps there could be a more descriptive name for this than `indexes`?

guenp: Good call! Also `indexes` should be renamed to `indices` at the very least :)

samples/error-correction/syndrome/Syndrome.qs
cgranade: Especially since this is a sample rather than a library, I suggest going with whatever is clearest to explain and provides the most insight to the reader; if that's `(Bool[], Pauli[], Int[])` rather than `(Bool, Pauli, Int)[]`, then I'm all for it. My only suggestion would be to add a comment explicitly to that effect in the API documentation above. Even for a sample, documentation comments are used by intelliSense and IQ#'s `?` command, and thus can be useful for communicating with readers.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+            for (index in indexes) {
+                ControlledPauli(encoding_bases[index], auxiliary, block[index]);
+            }
+            let auxiliary_result = Measure([PauliX], [auxiliary]);
+            let data_result = ForEach(BasisMeasure, Zip(encoding_bases, block));
+            Reset(auxiliary);
+            return ( auxiliary_result, data_result );
+        }
+    }
+}
```
cgranade, line 94:
```suggestion
}

```

samples/error-correction/syndrome/syndrome.py
cgranade: If you want to get really cute, you could do something like `paulis = list(map(qsharp.Pauli, range(1, 4)))`.

(I don't really recommend that.)

samples/error-correction/syndrome/syndrome.py
cgranade: You may find the `qsharp.Pauli.sample()` method interesting: https://github.com/microsoft/iqsharp/blob/master/src/Python/qsharp-core/qsharp/types.py#L35

samples/error-correction/syndrome/syndrome.py
```diff
@@ -0,0 +1,38 @@
...
+    indexes = list(range(args.qubits))
+    np.random.shuffle(indexes)
+    input_values = [np.random.rand() > .5 for n in range(args.qubits)]
+    encoding_bases = [np.random.choice(paulis) for n in range(args.qubits)]
+    result = SamplePseudoSyndrome.simulate(
+        input_values=input_values,
+        encoding_bases=encoding_bases,
+        indexes=indexes)
+    ancilla, data = result
```
cgranade, line 34: Elsewhere you've used auxiliary; would it make sense to do so here as well?

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// ## result
+    /// Measurement result
+    operation BasisMeasure(basis_qubit: (Pauli, Qubit)): Result {
+        let (basis, qubit) = basis_qubit;
+        let result = Measure([basis], [qubit]);
+        Reset(qubit);
+        return result;
+    }
+
+    operation Prepare(qubit: Qubit, value: Bool, basis: Pauli): Unit {
```
cgranade, line 52: I think the body is fairly self-explanatory, but an API documentation comment may help new readers nonetheless; I could see it going either way in this case.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// X basis we then project that phase difference onto the measurement basis.
+    ///
+    /// # Input
+    /// ## input_values
+    /// List of boolean values for data qubits
+    /// ## encoding_bases
+    /// List of Pauli bases
+    /// ## indexes
+    /// List of indexes on which to apply controlled pauli operators
+    operation SamplePseudoSyndrome (
```
cgranade, line 75: An `/// # Output` section listing what the expected return values are may help here.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+            input_values: Bool[],
+            encoding_bases: Pauli[], 
+            indexes: Int[]
+    ): ( Result, Result[] ) {
+        using ((block, auxiliary) = (Qubit[Length(input_values)], Qubit())) {
+            for ((qubit, value, basis) in Zip3(block, input_values, encoding_bases)) {
+                Prepare(qubit, value, basis);
+            }
+            H(auxiliary);
+            for (index in indexes) {
```
cgranade, line 85: You may be able to use `Zip` here as well.

samples/error-correction/syndrome/Syndrome.qs
cgranade: It may be simplest to not reset at all, since all qubits are measured immediately before releasing. That takes advantage of new 0.12 functionality, where resets are automatically added if a qubit has been measured right before it's deallocated (coherent unpreparation is still not automatic, though).

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Samples.Hardware.Syndrome
```
cgranade, line 3: Given that this is no longer specific to hardware (though it should also work fine on hardware, of course), would it make sense to drop that from the namespace here?

samples/error-correction/syndrome/Syndrome.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.Quantum.Sdk/0.12.20070124">
```
msoeken, line 1: Would this work with version `0.12.20072031`?

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// Apply a controlled Pauli operation for a given basis and control/target qubits
+    ///
+    /// # Input
+    /// ## basis
+    /// Pauli basis in which to apply the controlled pauli
+    /// ## control
+    /// Control qubit
+    /// ## target
+    /// Target qubit
+    operation ControlledPauli(basis: Pauli, control: Qubit, target: Qubit): Unit {
```
msoeken, line 23: Could this be implemented as `Controlled ApplyPauli(control, ([basis], [target]));` instead of using the three `if` statements?

samples/error-correction/syndrome/Syndrome.qs
efratshabtai: You might want to add a comment that the input arrays are expected to be equal in size.

efratshabtai:
Can this sample run on IonQ or Honeywell?
What will be the result in that case?

samples/error-correction/syndrome/syndrome.py
guenp: It looks like the above suggestions don't work because of a bug in `iqsharp`. I've filed a ticket here: https://github.com/microsoft/iqsharp/issues/256

samples/error-correction/syndrome/Syndrome.qs
guenp: Sure, I'll change it to `ErrorCorrection`

samples/error-correction/syndrome/Syndrome.qs
guenp: That's neat, thanks for the suggestion!

samples/error-correction/syndrome/syndrome.py
rmshaffer: Good catch @guenp, thanks for finding and filing that!

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+        if (PauliY == basis) {
+            CY(control, target);
+        }
+    }
+
+    /// # Summary
+    /// Measure qubit in a given basis and return the result
+    ///
+    /// # Input
+    /// ## basis_qubit
```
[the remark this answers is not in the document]
guenp: > That won't have any effect on callers, due to singleton–tuple equivalence

That's good to know, thanks for pointing this out!

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// Tuple of the Pauli basis and qubit to measure
+    /// 
+    /// # Output
+    /// ## result
+    /// Measurement result
+    operation BasisMeasure(basis_qubit: (Pauli, Qubit)): Result {
```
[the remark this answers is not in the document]
guenp: Yeah I've seen it come by several times, so I agree that'd be a good idea.

samples/error-correction/syndrome/Syndrome.qs
guenp: Sorry, I missed documenting this one.

guenp: By the way, this may also be a function we want to add to the standard libraries. What do you think?

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,94 @@
...
+    /// Creates a Pseudo Syndrome by using an auxiliary qubit.
+    /// This algorithm relies on several controlled pauli operations. When there is
+    /// no error introduced after state preparation, the circuit is trivial and no 
+    /// change is measured on the auxiliary qubit. However, if there are miscellaneous
+    /// rotations on the data qubits due to noise, due to phase kickback the auxiliary
+    /// qubit will gain a small phase shift. By measuring the auxiliary qubit in the
+    /// X basis we then project that phase difference onto the measurement basis.
+    ///
+    /// # Input
+    /// ## input_values
```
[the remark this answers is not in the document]
guenp: Thanks for catching that! I had planned to go through the code and catch this kind of stuff before converting from draft to PR but I appreciate your effort in pointing this out :)

guenp:
Thanks @cgranade @msoeken @efratshabtai  @rmshaffer  for the thorough code review! The purpose of the draft was really to discuss if we need to extend the sample to include random simulated errors (as to make the Syndrome usage more clear), however it looks like we prefer to leave the sample as-is and revisit after we make a decision on implementing noisy simulators. I'll convert this to a PR now, would appreciate a final pass from one of you.
@cgranade based on your comments it looks like there are some additions to make to the standard libraries. Where do you suggest to follow up on this? would this be a topic for our api review meeting or shall I just create a GH issue?

guenp:
> Can this sample run on IonQ or Honeywell?
What will be the result in that case?

@efratshabtai we should be able to run this on hardware depending on some pieces of the stack that are currently in development. The expected result is a measurable error on the syndrome, depending on the noise in the system.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Samples.ErrorCorrection.Syndrome
+{
```
cgranade, line 4:
```suggestion

namespace Microsoft.Quantum.Samples.ErrorCorrection.Syndrome {
```

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,105 @@
...
+    /// # Input
+    /// ## basis
+    /// Pauli basis in which to perform measurement
+    /// ## qubit
+    /// Qubit to measure
+    /// 
+    /// # Output
+    /// ## result
+    /// Measurement result
+    operation BasisMeasure(basis: Pauli, qubit: Qubit): Result {
```
cgranade, line 25: I'd suggest `MeasureInBasis` to match `PrepareInBasis` below. As a minor nit, we've generally used spaces between identifiers and their types as well:
```suggestion
    operation BasisMeasure(basis : Pauli, qubit : Qubit): Result {
```

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,105 @@
...
+    /// Prepare qubit in a given basis. Optionally flip the qubit if the value is True (One).
+    ///
+    /// # Input
+    /// ## qubit
+    /// Qubit to prepare
+    /// ## value
+    /// Value to prepare the qubit in (True for One, False for Zero)
+    /// ## basis
+    /// Basis to prepare the qubit in
+    operation PrepareInBasis(qubit: Qubit, value: Bool, basis: Pauli): Unit {
```
cgranade, line 40: The order of inputs here is different from the ordering used in `BasisMeasure`, such that I'd recommend using a common ordering for both. In particular, the [API principles](https://docs.microsoft.com/quantum/resources/contributing/api-design-principles) suggest ordering inputs based on commonality of partial application, which is almost always classical inputs before quantum. In this case, that'd point to `value : Bool, basis : Pauli, qubit : Qubit`, I think.

While the API principles don't apply to samples per se, looking at them can help make sure that users learning from your sample can transfer that knowledge to other parts of the Quantum Development Kit efficiently.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,105 @@
...
+    ): ( Result, Result[] ) {
+        // Check that input lists are of equal length
+        if ((Length(inputValues) != Length(encodingBases)) 
+            or (Length(inputValues) != Length(qubitIndices))) {
+            fail "Lengths of input values, encoding bases and qubitIndices must be 
+            equal. Found lengths: " 
+            + IntAsString(Length(inputValues)) + ", " 
+            + IntAsString(Length(encodingBases)) + ", " 
+            + IntAsString(Length(qubitIndices)) + ".";
+        }
```
cgranade, line 85: EqualityFactI may be useful here, as it prints out the expected and actual values in the case of failures. The other suggestion that I'd make is that string interpolation can avoid having to call `IntAsString` explicitly; i.e.: `$"Found lengths: {Length(inputValues)}, {Length(encodingBases)}, {Length(qubitIndices)}"`.

samples/error-correction/syndrome/Syndrome.qs
```diff
@@ -0,0 +1,105 @@
...
+
+        using ((block, auxiliary) = (Qubit[Length(inputValues)], Qubit())) {
+            for ((qubit, value, basis) in Zip3(block, inputValues, encodingBases)) {
+                PrepareInBasis(qubit, value, basis);
+            }
+            H(auxiliary);
+            // Apply Controlled Pauli's to data qubits, resulting in a phase kickback
```
cgranade, line 92:
```suggestion
            // Apply Controlled Pauli operations to data qubits, resulting in a phase kickback 
```

samples/error-correction/syndrome/Syndrome.qs
guenp: Thanks, I hadn't realized that I forgot to do this!

guenp:
Thanks, @cgranade ! I've implemented your suggestions. I would appreciate a read of the [README](https://github.com/microsoft/Quantum/pull/399/files?short_path=0b5db03#diff-0b5db03e28e0e284d6252cb30c41c01a) when you have a moment.

PR #405, opened 2020-08-17
Adding pre-requisites and fixing titles

geduardo:
Added a pre-requisite for the QML sample with Python visualization since you need to have installed Numpy and Matplotlib. Also modified the titles since the Samples browser for some reason doesn`t escape 'Q\#'.

Edit: I added an `environment.yml` file for the sample that uses Numpy and Matplotlib and added a short instruction of how to use it.  I also removed the lines loading the packages in Python, since they were outdated and they are no longer needed with the new conda package.


samples/machine-learning/half-moons/README.md
```diff
@@ -18,6 +18,7 @@ The model is trained on a half-moons dataset, loaded in C# using the System.Text
 ## Prerequisites
 
 - The Microsoft [Quantum Development Kit](https://docs.microsoft.com/quantum/install-guide/).
+- To run the Python version you also will need to install the Python libraries `numpy` and `matplotlib`. You can install them from the command line by typing `pip install <library name>`.
```
cgranade, line 21: This is a good point; the samples repo contains an environment.yml file that should work for all Python-based samples, but that gets lost when onboarding a particular sample to docs.microsoft.com/samples. It may be worth making a copy of that environment.yml file here as well to automate installing NumPy and Matplotlib. I'm a little wary of telling people to `pip install` if our golden path for Python installation is to use `conda`.

geduardo: Do you mean to copy [this file](https://github.com/microsoft/Quantum/blob/geduardo/adding-prerequisites/samples/interoperability/python/environment.yml)? 

Another option could be to add Numpy and Matplotlib to the default environment created here: https://github.com/microsoft/Quantum/blob/geduardo/adding-prerequisites/samples/interoperability/python/environment.yml

samples/machine-learning/half-moons/host.py
```diff
@@ -10,8 +10,6 @@
 plt.style.use('ggplot')
 
 import qsharp
-qsharp.packages.add("Microsoft.Quantum.MachineLearning::0.11.2004.2825")
-qsharp.reload()
```
rmshaffer, line 14: Just wanted to call out that the removal of these lines won't work properly until the upcoming QDK release has completed (which includes the support for reading the packages from the `.csproj`). ETA for that release is August 24. (That said, I think this wouldn't work on the most recent releases _anyway_, since it references an 0.11.* version.)

geduardo: Yes, I thought about it. But since it doesn't work anyways... how bad is to have it unusable until the 24th? 

What it is important is that this is the sample that we are going to have for when the video is released, since this procedure is what I'm going to record.

rmshaffer: Yes, that makes sense to me. I suspect it's probably fine, but I'll let @cgranade make the call as the owner of these samples.

cgranade: In general, if the sample is broken until release, it's best to either defer merging the PR, or targeting a release branch. I agree it's not the end of the world given its five days until the sample is fixed, but we do want to make sure people have a good first impression of samples since that's a primary entry point to the Quantum Development Kit.

rmshaffer: That makes sense, @cgranade. It looks like the sample may already be broken in its current state (since it's referencing an old 0.11.* version of the package), so perhaps updating that version would be a good immediate step for this PR?

@geduardo, I can take the action item to come update these samples for the new functionality after the next release is complete.

cgranade: Now that we're on the other side of the release, I think we can resume this PR and use the new automatic package loading functionality introduced in the August. Thoughts, @geduardo and @rmshaffer?

samples/machine-learning/half-moons/README.md
cgranade: The environment.yml for samples/interoperability/python is fairly specialized to the process tomography case shown there, such that the environment.yml file can be a lot simpler in cases like this one, such as the one you added with a1656dc7a7bd0331e2064f009c191bd8eb7f1d47.

samples/machine-learning/half-moons/host.py
geduardo: True, I almost forgot about this PR. What new functionality was added? 

Relevant information: A video showing how to run these samples is being produced. Changes in the process to run the samples should be avoided, or if it needs to be changed it should be done ASAP.

cgranade: Automatic package loading was added in the previous release, so that the extraneous "fail" messages at the console can be avoided by including a csproj file with the relevant packages in the same folder as the Q# notebook.

```
<Project Sdk="Microsoft.Quantum.Sdk/0.12.20082513">
    <PropertyGroup>
        <OutputType>Library</OutputType>
        <TargetFramework>netstandard2.1</TargetFramework>
        <IQSharpLoadAutomatically>true</IQSharpLoadAutomatically>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Microsoft.Quantum.MachineLearning" Version="0.12.20082513" />
    </ItemGroup>
</Project>
```

(example adapted from https://docs.microsoft.com/quantum/user-guide/host-programs?tabs=tabid-python#using-q-code-from-other-projects-or-packages-1)

Given that this would significantly improve how the sample looks, I'd suggest getting it in ASAP if you and @KittyYeungQ agree.

rmshaffer: @geduardo @cgranade I have the changes for automatic package loading available in PR https://github.com/microsoft/Quantum/pull/411. I'd be happy to merge that one; I believe it's ready.

cgranade: Reviewing now, then. Thanks!

geduardo: @rmshaffer  @cgranade Let me get this clear. How does this automatic package loading affect this PR since using the Conda environment packages were already loaded correctly?  If I understand it correctly, we still need the Conda environment since we are loading Python packages like Numpy or Matplotlib.

rmshaffer: @geduardo Yes, sorry, "package" is overloaded here. 😊 This "automatic package loading" feature is not for the Python packages, but for the NuGet packages like `Microsoft.Quantum.MachineLearning`. You can see the relevant changes in PR #411. The only difference from what you have here is that it also adds `<IQSharpLoadAutomatically>true</IQSharpLoadAutomatically>` to the .csproj files.

geduardo: Ok, I see, then I think the video doesn't need to be modified @KittyYeungQ. Once one of the two PRs gets merged, the other one will get some small conflicts I think. You can merge #411 and then I'll resolve the conflicts here.

geduardo: I was wrong! No conflicts since both implement the same changes in the "conflicting" files 😆 Thanks!

PR #410, opened 2020-08-20
Execution path visualizer integration w/ state visualizer

theRoughCode:
This PR introduces the long-awaited execution path visualizer into the state visualizer. This includes a pretty major overhaul of the existing code as the advance/wait commands are no longer needed since we are using the circuit to jump around the program. As such, maybe a different sample adjacent to the existing one might be best.

Here are its features:
1. On page load, the given Q# program (e.g. `SimpleGrover.SearchForMarkedInput`) is simulated by the Quantum simulator to which we've attached on the `ExecutionPathTracer` and also the `StateVisualizer` (as before).
2. The program is let to run to fully execute/simulate as opposed to the previous version where the program would pause execution to wait for the user to advance to the next step. After the simulation is finished, the visualization of the execution path is rendered on-screen together with the initial state in the chart.
3. The user is then able to click on individual gates to step through their program.
4. Additionally, we add the capability to zoom in/out of operations to give the user more fine-grained control of their debugging/learning experience.

This sample uses:
1. The `ExecutionPathTracer` in IQ#. Specifically, the one introduced in [this PR](https://github.com/microsoft/iqsharp/pull/280).
2. The `Sqore` package (found [here](https://dev.azure.com/ms-hiddenfalls-interns-20/_git/KohRaphael2020)) for visualization which will be released as a standalone npm package (either internally or open-source).


theRoughCode:
Updated base branch to a feature branch instead as this PR introduces a major overhaul to existing code.

guenp:
@theRoughCode just wanted to check in with you, if there are any blockers left to merging this PR? Is sqore still something that needs to be installed manually?

theRoughCode:
@guenp I don't think this app works yet. We'll still need to install sqore manually. It's also been a while since I've checked on the progress of the library, so there could've been developments for open-sourcing the app.

I believe we've been blocked on finding a suitable name for the sqore package for open-source.  However, I haven't heard about any updates since October. Maybe @anpaz or @rmshaffer might know better?

guenp:
Thanks @theRoughCode for the update! Let me check back with the team regarding the naming, so we can get this PR unblocked.

PR #418, opened 2020-08-25
Update C# callable implementations to use new naming conventions

samarsha:
This updates the Quantum repo for the breaking changes to the Q# runtime that will be made by microsoft/qsharp-runtime#349:

* Wrap C# callable member names in underscores (e.g., `Body` -> `__Body__`, `Init` -> `__Init__`).

This shouldn't be merged until microsoft/qsharp-runtime#349 is merged.

### DIFF
--- a/samples/simulation/h2/command-line/Host.cs
+++ b/samples/simulation/h2/command-line/Host.cs
@@ -3,39 +3,23 @@
 
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 
-using Microsoft.Quantum.Simulation.Core;
 using Microsoft.Quantum.Simulation.Simulators;
 
 namespace Microsoft.Quantum.Samples.H2Simulation
 {
     class Program
     {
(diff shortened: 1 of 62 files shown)


samples/simulation/h2/command-line/Host.cs
```diff
@@ -35,7 +35,7 @@ static void Main(string[] args)
 
                 // To call a Q# operation that takes unit `()` as its input, we need to grab
                 // the QVoid.Instance value.
-                var bondLengths = H2BondLengths.Body.Invoke(QVoid.Instance);
+                var bondLengths = H2BondLengths.__Body__.Invoke(QVoid.Instance);
```
bettinaheim, line 38: It would be better to replace these with the sim.Run invocations - any reason why this should be invoked this way?

samarsha: I don't think `sim.Run` works here, but I updated the sample to use `op.Run(sim, ...)`.

msoeken (approved):

PR #423, opened 2020-09-11
Replace references to branch master with main

anpaz-msft:


samples/chemistry/IntegralData/Broombridge_v0.2/H2_sto-3g.yaml
```diff
@@ -1,4 +1,4 @@
-"$schema": https://raw.githubusercontent.com/Microsoft/Quantum/master/Chemistry/Schema/broombridge-0.2.schema.json
```
cgranade, line 1: These changes raise an interesting point in that schema URLs are generally required to be stable. Will the contents of the schema still be available using the old URL? Should we create an aka.ms link that we can redirect if we need to move things again in the future?

cgranade (approved):
Looks good to me, modulo one quick thought with schema URLs.

samples/chemistry/IntegralData/Broombridge_v0.2/H2_sto-3g.yaml
anpaz-msft: I'm hopeful that redirects to default branch, as described [here](https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/) can take care of this.

PR #431, opened 2020-09-22
Guenp/20147 simple rus

guenp:
Simple RUS sample (figure 8 from https://arxiv.org/abs/1311.1074)


samples/algorithms/repeat-until-success/SimpleRUS.qs
```diff
@@ -0,0 +1,121 @@
...
+            // Assert valid starting states for all qubits
+            AssertMeasurement([PauliZ], [register[0]], Zero,
+             "Auxiliary qubit is not in |0⟩ state.");
+            AssertQubitIsInState(register[1], inputBasis, inputValue);
+            
+            if (ApplyAndMeasureSimpleRUSCircuit(register) == Zero) { //|0⟩
+                set success = true;
+            }
+
+            set success = success or (numIter >= limit);
```
swernli, line 93: In the V-gate RUS, you use a second variable to capture the loop exit condition, which I think you should do here too, that way hitting loop limit does not get reported as a success.

samples/algorithms/repeat-until-success/SimpleRUS.qs
```diff
@@ -0,0 +1,121 @@
...
+            set numIter = numIter + 1;
+            Reset(register[0]);
```
swernli, line 95: This Reset can be moved into and else scope under `if (ApplyAndMeasureSimpleRUSCircuit(register)) == Zero)` and at that point can be replaced with `X` instead. Alternatively, since the auxiliary qubit state is not needed for the rerun of the circuit, the call to `Measure` inside of `ApplyAndMeasureSimpleRUSCircuit` can be replaced with `MResetZ` and pull all the conditional reset logic into there.

samples/algorithms/repeat-until-success/SimpleRUS.qs
```diff
@@ -0,0 +1,121 @@
...
+     : Result {
+        H(register[0]);
+        T(register[0]);
+        CNOT(register[0], register[1]);
+        H(register[0]);
+        CNOT(register[0], register[1]);
+        T(register[0]);
+        H(register[0]);
+
+        return Measure([PauliZ], [register[0]]);
```
swernli, line 119: See comment above; I think this could reasonably be a call to `MResetZ` instead, but if you choose to keep the reset out of this operation then I would be in favor of replacing this with `M` for simplicity.

swernli (approved):
This is great, thanks for working on this! I have some feedback, but over all I think this will fit our needs well.

samples/algorithms/repeat-until-success/SimpleRUS.qs
```diff
@@ -0,0 +1,121 @@
...
+        // Initialize results to One by default.
+        mutable success = false;
+        mutable numIter = 0;
+        if (inputValue) {
+            // Prepare target qubit in |0⟩ or |1⟩ state, depending on input value
+            if (inputValue) {
+                X(register[1]);
+            }
+            PrepareQubit(inputBasis, register[1]);
+        }
```
swernli, line 81: Whoops! I just noticed this double if, which has the side effect of always preparing Zero in the Pauli-Z basis instead of the supplied basis.
```suggestion
        // Prepare target qubit in |0⟩ or |1⟩ state, depending on input value
        if (inputValue) {
            X(register[1]);
        }
        PrepareQubit(inputBasis, register[1]);
```

guenp: Whoops! Thanks, great catch.

samples/algorithms/repeat-until-success/SimpleRUS.qs
guenp: Ah I see what you mean, good point. I'll revert this change.

PR #435, opened 2020-10-06
Reduce warnings and improve style guide adherance.

cgranade:
This PR reduces the number of warnings by updating to avoid deprecated functions and operations. This PR also makes improvements to readability by using recent improvements in the libraries and by improving style guide adherance.


cgranade:
It looks like whatever transient issue was affecting builds has since been resolved, such that this should now be reviewable.

msoeken (approved):
Thanks for all that work @cgranade. Nice to also see the evolution of the libraries in these changes.

PR #439, opened 2020-11-04
Temporary Title

ScottCarda-MS:


guenp:
@ScottCarda-MS , will the samples still run without the `EntryPoint` definitions?

ScottCarda-MS:
> @ScottCarda-MS , will the samples still run without the `EntryPoint` definitions?

I'm not totally sure yet. The ones that have C# drivers should still run I believe, but we may need to do something else for the ones without C# drivers.

guenp:
@ScottCarda-MS Hi Scott, just wanted to check in with you if this is still a work in progress?

ScottCarda-MS:
> @ScottCarda-MS Hi Scott, just wanted to check in with you if this is still a work in progress?

Yep, it is just taking a bit longer than we thought to be able to refactor this repo. There were some feature requests made in the effort to do so, including the Multiple Entry Points feature that I have some PRs up for now. These features were requested to allow for a more proper refactor of the Samples repo. This is still in my work stack 😄.

PR #442, opened 2020-11-11
Restore valid base image version

rmshaffer:
Looks like #437 contained an inadvertent change to `Dockerfile` that broke the Docker image build.

e.g., from https://github.com/microsoft/Quantum/runs/1383424059:
```
Step 1/10 : FROM mcr.microsoft.com/quantum/iqsharp-base:0.13.201118141-beta
manifest for mcr.microsoft.com/quantum/iqsharp-base:0.13.201118141-beta not found: manifest unknown: manifest tagged by "0.13.201118141-beta" is not found
Error response from daemon: No such image: public/quantum/samples:77b68faa38df8a601ba3fcc1ffef3094be242cba
```

This change reverts back to use the previously-released version 0.13.20102604.

### DIFF
--- a/Build/images/samples/Dockerfile
+++ b/Build/images/samples/Dockerfile
@@ -2,7 +2,7 @@
 # https://github.com/microsoft/iqsharp/blob/main/images/iqsharp-base/Dockerfile.
 # As per Binder documentation, we choose to use an SHA sum here instead of a
 # tag.
-FROM mcr.microsoft.com/quantum/iqsharp-base:0.13.201118141-beta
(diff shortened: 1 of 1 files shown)


cgranade:
That's my bad, sorry; I should have caught that beta Docker images aren't actually published. Thanks for catching this!

cgranade (approved):

rmshaffer:
> That's my bad, sorry; I should have caught that beta Docker images aren't actually published. Thanks for catching this!

No worries at all - that's why we have CI to catch these things!

PR #443, opened 2020-11-12
Update Quantum SDK references to enable local e2e builds after compiler breaking change

cesarzc:
This change updates references to the Quantum SDK such to make them compatible with the latest version of the compiler and enable local e2e builds.


anpaz (approved):

cesarzc:
Will wait for the November release to update references.

Issue #376, opened 2020-06-24
Port order-finding sample to Q#


msoeken:
Currently, the order-finding sample uses a C# host. This can be changed to a Q# host.

manav28:
Could I take this one?

cgranade:
@manav28: If you're interested, we'd definitely love a contribution to this one! A good place to start could be looking at the [contribution guidelines for code](https://docs.microsoft.com/quantum/resources/contributing/code), but we're also happy to answer questions here as well. Thank you!

manav28:
Are there any examples of a Q# host for reference? Are we porting the Q# host to a .qs file or a jupyter notebook?

cgranade:
Most of the samples have been modified to use Q# standalone programs, as detailed in the [command-line part of the user guide](https://docs.microsoft.com/quantum/user-guide/host-programs?tabs=tabid-python#q-from-the-command-line). Happy to help if you have any questions, though, and we greatly appreciate the contribution!

cgranade:
This should be closed thanks to #392. Thank you so much, @manav28!

Issue #375, opened 2020-06-22
RippleCarryAdderCDKM not releasing qubits in zero state on completion


apm04:
I have converted the Jupyter notebook AdderExample.ipynb from 
https://github.com/microsoft/Quantum/tree/master/samples/arithmetic
to a standalone program with a host.py and TestAdd.qs. (uploaded, remove '.txt')

[host.py.txt](https://github.com/microsoft/Quantum/files/4815384/host.py.txt)
[TestAdd.qs.txt](https://github.com/microsoft/Quantum/files/4815385/TestAdd.qs.txt)

The adder works as expected with RippleCarryAdderTTK and RippleCarryAdderD.

With RippleCarryAdderCDKM the following error message is given:

Unhandled exception. Microsoft.Quantum.Simulation.Simulators.Exceptions.ReleasedQubitsAreNotInZeroState: Released qubits are not in zero state.

It appears the implementation of RippleCarryAdderCDKM is not releasing all internally used qubits on completion. Curiously the error does not appear using the Jupyter notebook itself.

**System information**
 - OS: Ubuntu 18.04
 - Python 3.6.9
 - Visual Studio Code 1.46.1
 - MS Quantum Development Kit for Visual Studio Code 0.11.20060403

cgranade:
Thanks for raising this, I believe that this should be addressed by https://github.com/microsoft/QuantumLibraries/pull/303, but please let us know if the issue persists for you. Thank you!

PR #322, opened 2020-02-26
Address feedback on QML samples.

cgranade:
This PR applies feedback from #321.

Issue #212, opened 2019-07-19
Reference to unknown namespace Microsoft.Quantum.Intrinsic


stefante:
**Describe the bug**
I am following the instructions for the "hello world"program

namespace Quantum.Bell {
    open Microsoft.Quantum.Intrinsic;
    open Microsoft.Quantum.Canon;

    operation HelloQ () : Unit {
        Message("Hello quantum world!");
    }
}

on "dotnet run" I always get the same error message 

"Reference to unknown namespace Microsoft.Quantum.Intrinsic" 

How do I properly reference this namespace? Thanks

anpaz-msft:
Hi @stefante, can you add more details about the error message? There is typically a code and a line number associated with the error that will give us more information.
Also, can you let us know what version of .net core (`dotnet --info`) and what OS this is running on?

stefante:
Thanks for the reply and your help, @anpaz-msft 

I am running **dotnet 2.2.301**  and **Mac OS Mojave 10.14.5**

**The first messages are:** 

- Bell.qs(1,23): error QS0001: The namespace body is invalid [/Users....../Bell/Bell.csproj]

- Bell.qs(3,5): error QS0001: Reference to unknown namespace Microsoft.Quantum.Intrinsic [/Users..... /Bell/Bell.csproj]

**Bell.csproj was autogenerated and includes the 2 following package references**

- PackageReference Include="Microsoft.Quantum.Canon" Version="0.2.1806.3001-preview" 
- PackageReference Include="Microsoft.Quantum.Development.Kit" Version="0.2.1806.3001-preview" /

cgranade:
@stefante: It sounds like your project templates are out-of-date, as we're currently on [version 0.8.1907.1701](https://www.nuget.org/packages/Microsoft.Quantum.Development.Kit/0.8.1907.1701). In particular, the namespace name `Microsoft.Quantum.Intrinsic` was introduced with Q# version 0.6, such that this error makes sense when using a package reference to version 0.2 of the Quantum Development Kit.

To update, I'd suggest two steps:

### Update your project templates to version 0.8 ###

```sh
dotnet new -i Microsoft.Quantum.ProjectTemplates::0.8.1907.1701
```

### Migrate your existing `Bell.csproj` to use 0.8 ###

[...]

PR #444, opened 2020-11-25
Update to QDK release 0.14.2011.120240

guenp:
Removed optional parameters.
@msoeken @cgranade are there any changes we should make to the samples based on below API improvements?

> Updated the Microsoft.Quantum.Preparation namespace to be more consistent with style guide and API design principles, and to support purified mixed states with additional data (see proposal, review notes and PRs #212, #322, #375, #376).


msoeken (approved):
Looks good to me. Thanks @guenp

PR #449, opened 2020-12-29
Updated packages of Lithdemo and removed Windows dependency

RolfHuisman:
Updated packages because of a low security issue in init (dependabot #446).
Trying to test the application on mac highlighted a windows style path, so changed that one to an os invariant one.


cgranade:
Thanks for fixing this, @RolfHuisman! Looks great to me!

cgranade (approved):

PR #454, opened 2021-01-14
Started updating to 0.14.210121247-alpha.

cgranade:
This PR updates the samples to use QEP 1 (`use` statements instead of `using` blocks), introduced with 0.14.210121247-alpha. So far, this PR is in draft as many samples still need updated, and as no IQ# notebooks have yet been updated.


guenp:
@cgranade OK if we close this in favor of upgrading to `0.15.2101125897`?

cgranade:
> @cgranade OK if we rescope this in favor of upgrading to `0.15.2101125897`?

Sounds good, yeah. The 0.14-beta should be entirely compatible with 0.15.2101125897, so that for the most part that rescoping should just be updating version numbers again.

samples/algorithms/chsh-game/CHSHGame.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.Quantum.Sdk/0.14.2011120240">
+<Project Sdk="Microsoft.Quantum.Sdk/0.14.210121247-alpha">
```
msoeken: Should this be updated to 0.15?

msoeken (approved):

PR #455, opened 2021-01-22
Adjust override sample to account for ongoing refactoring.

swernli:
See https://github.com/microsoft/qsharp-runtime/pull/476

This is required to unblock the PR above. The `QSimM` class is gone, so directly overriding the intrinsic method is now correct. That still works with the current simulator design, so this change should be safe to merge even before the above goes into qsharp-runtime main branch.


swernli:
I need to do a little more validation before I'm ready to merge this PR, so I'll leave it in draft until I've confirmed the correct behavior using the e2e build, then publish and ping for review.

swernli:
I am going to find a different way to handle this in the refactor PR that should eliminate the need for consumers to make these changes. I did test that this change behaves as expected, so that will give us a path forward when we are ready to remove the old deprecated classes used in this sample

PR #459, opened 2021-01-28
Update Q# notebooks to 0.15 syntax.

cgranade:


samples/algorithms/database-search/Database Search.ipynb
```diff
@@ -326,7 +330,7 @@
    "file_extension": ".qs",
    "mimetype": "text/x-qsharp",
    "name": "qsharp",
-   "version": "0.10"
+   "version": "0.14"
```
msoeken, line 333: What is this version number?

msoeken (approved):
Looks good to me, thanks!

samples/algorithms/database-search/Database Search.ipynb
guenp: @msoeken I think this is metadata generated by Jupyter.

cgranade: Indeed, that metadata is generated from here: https://github.com/microsoft/iqsharp/blob/406629f9378aebefd1cd54e5877471c8d61cfbc6/src/Kernel/KernelProperties.cs#L28

That needs manually updated with version bumps, but that got missed during the 0.14 → 0.15 bump.

PR #448, opened 2020-12-22
Implement flame graph visualization

aman3014:
Addition of a visualize method to Program.cs
To generate a flame graph, follow these steps:

1. Run Program.cs with the `visualize` option by specifying the generator and the resource (more information in the help text) to be visualized. Save the output in a file.
2. Use [flamegraph.pl](https://github.com/brendangregg/FlameGraph#3-flamegraphpl) with the above file to generate an svg of the flame graph.

PR #467, opened 2021-02-15
Flamegraph visualization example in Shor's algorithm

msoeken:
This merges #448 and some auxiliary code to enable resource estimation in Shor's algorithm.

### DIFF
--- a/samples/algorithms/integer-factorization/FlameGraphCounter.cs
+++ b/samples/algorithms/integer-factorization/FlameGraphCounter.cs
@@ -5,29 +5,28 @@ using Microsoft.Quantum.Simulation.Core;
 using Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators;
 
 namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime {
-    public class FlameGraphCounter : IQCTraceSimulatorListener, ICallGraphStatistics {
-        class OperationCallRecord
-        {
-            public HashedString OperationName;
-            public OperationFunctor FunctorSpecialization;
-            public double[] GlobalCountersAtOperationStart;
-        }
-
+    public class FlameGraphCounter : IQCTraceSimulatorListener {
         private readonly PrimitiveOperationsCounterConfiguration configuration;
+        
+        /// <summary>
+        /// Contains all fully specified calls including resource count for the operation excluding all children.
+        /// </summary>
+        public Dictionary<string, double> FlameGraphData { get; } = new Dictionary<string, double>();
+
+        /// <summary>
+        /// Sets which resource to track (e.g., CNOT)
+        /// </summary>
+        public PrimitiveOperationsGroups ResourceToVisualize { get; set; } = PrimitiveOperationsGroups.CNOT;
 
         /// <summary>
         /// Number of Primitive operations performed since the beginning of the execution.
         /// Double type is used because all the collected statistics are of type double.
         /// </summary>
         private readonly double[] globalCounters;
-        private readonly Stack<OperationCallRecord> operationCallStack;
-        private Dictionary<string, double> flameGraphData;
+
+        private readonly Stack<double[]> operationCallStack;
+
         private string callStack = "";
-        public int resourceToVisualize {get; set;} = (int) PrimitiveOperationsGroups.CNOT;
-        private readonly int PrimitiveOperationsCount;
-        private readonly StatisticsCollector<CallGraphEdge> stats;
-        public PrimitiveOperationsCounterConfiguration GetConfigurationCopy() { return Utils.DeepClone(configuration); }
-        public IStatisticCollectorResults<CallGraphEdge> Results { get => stats as IStatisticCollectorResults<CallGraphEdge>; }
 
         /// <param name="statisticsToCollect">
         /// Statistics to be collected. If set to null, the
@@ -36,44 +35,30 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime {
         public FlameGraphCounter(PrimitiveOperationsCounterConfiguration config, IDoubleStatistic[]  statisticsToCollect = null )
         {
             configuration = Utils.DeepClone(config);
-            PrimitiveOperationsCount = configuration.primitiveOperationsNames.Length;
-            globalCounters = new double[PrimitiveOperationsCount];
-            operationCallStack = new Stack<OperationCallRecord>();
+            globalCounters = new double[configuration.primitiveOperationsNames.Length];
+            operationCallStack = new Stack<double[]>();
             AddToCallStack(CallGraphEdge.CallGraphRootHashed,OperationFunctor.Body);
-            stats = new StatisticsCollector<CallGraphEdge>(
-                config.primitiveOperationsNames,
-                statisticsToCollect ?? StatisticsCollector<CallGraphEdge>.DefaultStatistics()
-                );
-            this.flameGraphData = new Dictionary<string, double>();
-        }
-
-        public Dictionary<string, double> GetFlameGraphData() {
-            return new Dictionary<string, double>(this.flameGraphData);
         }
 
         // returns the resulting call stack after adding the given function call
-        private static string PushString(string s, string add) {
+        private static string PushString(string s, string add)
+        {
             Debug.Assert(s != null);
             Debug.Assert(add != null);
             return s.Length == 0 ? add : s + ";" + add;
         }
 
         // returns the remaining call stack after removing a function call from the top of the stack
-        private static string PopString(string s) {
+        private static string PopString(string s)
+        {
             int index = s.LastIndexOf(';');
             return index > 0 ? s.Substring(0, index) : "";
         }
 
         private void AddToCallStack( HashedString operationName, OperationFunctor functorSpecialization)
         {
-            operationCallStack.Push(
-                new OperationCallRecord()
-                {
-                    GlobalCountersAtOperationStart = new double[PrimitiveOperationsCount],
-                    OperationName = operationName,
-                    FunctorSpecialization = functorSpecialization
-                });
-            globalCounters.CopyTo(operationCallStack.Peek().GlobalCountersAtOperationStart, 0);
+            operationCallStack.Push(new double[globalCounters.Length]);
+            globalCounters.CopyTo(operationCallStack.Peek(), 0);
             callStack = PushString(callStack, operationName);
         }
 
@@ -111,7 +96,7 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime {
         /// </summary>
         public void OnPrimitiveOperation(int id, object[] qubitsTraceData, double PrimitiveOperationDuration)
         {
-            Debug.Assert(id < PrimitiveOperationsCount);
+            Debug.Assert(id < globalCounters.Length);
             globalCounters[id] += 1.0;
         }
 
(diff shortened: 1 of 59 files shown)


samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
```diff
@@ -0,0 +1,48 @@
...
+            var primitiveOperationsIdToNames = new Dictionary<int, string>();
+            Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.Utils.FillDictionaryForEnumNames<PrimitiveOperationsGroups, int>(primitiveOperationsIdToNames);
+            var cfg = new PrimitiveOperationsCounterConfiguration {
+                primitiveOperationsNames = primitiveOperationsIdToNames.Values.ToArray()
+            };
+            operationsCounter = new FlameGraphCounter(cfg);
+            this.tCoreConfig.Listeners.Add(operationsCounter);
+        }
+
+        public Dictionary<string, double> FlameGraphData => operationsCounter.FlameGraphData;
```
cgranade, line 46: Since Dictionary is a mutable data structure, this property allows callers to change the contents of operationsCounter's internal data; would it be worth either making a copy or returning this as an ImmutableDictionary?

samples/algorithms/integer-factorization/README.md
```diff
@@ -11,3 +11,27 @@ a part of Microsoft.Quantum.Canon library.
...
+
+1. Clone the [flamegraph repository](https://github.com/brendangregg/FlameGraph)
+2. Install perl
+3. Run program with the `visualize` command by specifying the generator and the resource (more information in the help text) to be visualized. Save the output in a file.
+4. Use flamegraph.pl with the above file to generate an svg of the flame graph.
+
+Example usage:
+```
+dotnet run -- visualize -g 4 -r 0
+cp output.txt [the flamegraph directory]
```
cgranade, line 29: Where is output.txt set as the default? I was having trouble finding that above.

samples/algorithms/integer-factorization/README.md
```diff
@@ -11,3 +11,27 @@ a part of Microsoft.Quantum.Canon library.
 - [Program.cs](./Program.cs): C# console application running Shor's algorithm
   on Quantum simulator
 - [IntegerFactorization.csproj](./IntegerFactorization.csproj): Main C# project for the sample.
+
+# Flame Graph Visualization #
+
+This sample also contains an adapter for the [ResourcesEstimator](https://github.com/microsoft/qsharp-runtime/tree/974a385cc57c2b663e8134c1f3170f9cb8ae5fb1/src/Simulation/Simulators/ResourcesEstimator) allowing it to produce a resource utilization stack trace which can then be used to produce a flame graph. More details can be found in [this article](https://aman3014.medium.com/flame-graphs-for-q-f4f9bb076d88).
```
cgranade, line 17: Would it help to link to the docs instead of the source code here?

samples/algorithms/integer-factorization/README.md
```diff
@@ -11,3 +11,27 @@ a part of Microsoft.Quantum.Canon library.
...
+
+To generate a flame graph, follow these steps:
+
+1. Clone the [flamegraph repository](https://github.com/brendangregg/FlameGraph)
+2. Install perl
```
cgranade, line 22:
```suggestion
2. Install Perl
```
May be out of scope, but even a pointer as to how to do so may be helpful. In particular, I'd think that Perl is likely to come by default on most Linux and macOS systems, such that using chocolatey to install StrawberryPerl may be useful for Windows users.

Have you checked if this all works in the samples devcontainer for VS Code / GH Codespaces / mybinder.org usage?

samples/algorithms/integer-factorization/README.md
```diff
@@ -11,3 +11,27 @@ a part of Microsoft.Quantum.Canon library.
...
+```
+dotnet run -- visualize -g 4 -r 0
+cp output.txt [the flamegraph directory]
+cd [the flamegraph directory]
+perl flamegraph.pl output.txt > output.svg
+```
+
+## Manifest ##
+
+- [FlameGraphCounter.cs](./FlameGraphCounter.cs): Counter for the adapted resources estimator.
```
cgranade, line 36: Generally, manifest links should be absolute URLs so that they resolve correctly when imported to docs.microsoft.com/samples.

samples/algorithms/integer-factorization/FlameGraphCounter.cs
```diff
@@ -0,0 +1,98 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators;
+
+namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime {
+    public class FlameGraphCounter : IQCTraceSimulatorListener {
```
cgranade, line 8: Some comments here explaining what flamegraph.pl expects may help the reader interpret this sample; it wasn't clear to me, for instance, what the significance of `;` in call stack lines was until I looked at the flamegraph.pl docs.

samples/algorithms/integer-factorization/FlameGraphCounter.cs
```diff
@@ -0,0 +1,98 @@
...
+        private readonly PrimitiveOperationsCounterConfiguration configuration;
+        
+        /// <summary>
+        /// Contains all fully specified calls including resource count for the operation excluding all children.
+        /// </summary>
+        public Dictionary<string, double> FlameGraphData { get; } = new Dictionary<string, double>();
```
cgranade, line 14: Similarly here as below, this exposes a mutable data structure as readonly, such that callers can still change the contents of `FlameGraphData`; would it be worth using an ImmutableDictionary here?

samples/algorithms/integer-factorization/FlameGraphCounter.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>
+        /// Sets which resource to track (e.g., CNOT)
```
cgranade, line 17: Is this the only resource that shows up in the flamegraph, then?

samples/algorithms/integer-factorization/FlameGraphCounter.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            Debug.Assert(s != null);
+            Debug.Assert(add != null);
+            return s.Length == 0 ? add : s + ";" + add;
+        }
+
+        // returns the remaining call stack after removing a function call from the top of the stack
+        private static string PopString(string s)
+        {
+            int index = s.LastIndexOf(';');
```
cgranade, line 49: Elsewhere you've used `var` in this case?

samples/algorithms/integer-factorization/FlameGraphCounter.cs
```diff
@@ -0,0 +1,98 @@
...
+        public void OnOperationEnd(object[] returnedQubitsTraceData)
+        {
+            var globalCountersAtOperationStart = operationCallStack.Pop();
+            Debug.Assert(operationCallStack.Count != 0, "Operation call stack must never get empty");
+            double[] difference = Utils.ArrayDifference(globalCounters, globalCountersAtOperationStart);
+            globalCountersAtOperationStart.CopyTo(globalCounters, 0);
+            if (FlameGraphData.ContainsKey(callStack))
+                FlameGraphData[callStack] += difference[(int)ResourceToVisualize];
+            else
+                FlameGraphData.Add(callStack, difference[(int)ResourceToVisualize]);
```
cgranade, line 81: I recognize braces are optional here, but I would suggest that they still improve readability in this case.
```suggestion
            if (FlameGraphData.ContainsKey(callStack))
            {
                FlameGraphData[callStack] += difference[(int)ResourceToVisualize];
            }
            else
            {
                FlameGraphData.Add(callStack, difference[(int)ResourceToVisualize]);
            }
```

cgranade (changes requested):
I think this is looking pretty good — marked as request changes mostly to make sure we don't forget to add copyright headers.

samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq;
+using System.Collections.Generic;
+
+using Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime;
+using Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators;
+
+namespace Microsoft.Quantum.Simulation.Simulators {
+    public class FlameGraphResourcesEstimator : ResourcesEstimator {
+        private FlameGraphCounter operationsCounter;
+        public static new QCTraceSimulatorConfiguration RecommendedConfig() =>
```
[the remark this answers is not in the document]
msoeken: Agree, but in this particular case it's to be consistent with the `RecommendedConfig` method in `ResourcesEstimator`.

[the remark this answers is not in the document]
cgranade: Ah, got it. That makes sense; may even be good to call that out with a comment? I should have noticed the `new` keyword for that...

samples/algorithms/integer-factorization/FlameGraphCounter.cs
msoeken: I understand the point, but would avoid the copying in this case, since the data in the dictionary can become very large.

samples/algorithms/integer-factorization/FlameGraphCounter.cs
msoeken: Yes

samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
```diff
@@ -0,0 +1,48 @@
...
+        public FlameGraphResourcesEstimator(QCTraceSimulatorConfiguration config, PrimitiveOperationsGroups resourceToVisualize = PrimitiveOperationsGroups.CNOT) : base(WithoutPrimitiveOperationsCounter(config)) {
+            operationsCounter.ResourceToVisualize = resourceToVisualize;
+        }
+
+        private static QCTraceSimulatorConfiguration WithoutPrimitiveOperationsCounter(QCTraceSimulatorConfiguration config) {
+            config.UsePrimitiveOperationsCounter = false;
+            return config;
+        }
+ 
+        protected override void InitializeQCTracerCoreListeners(IList<IQCTraceSimulatorListener> listeners) {
```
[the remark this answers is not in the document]
msoeken: Yes, it's `IList` in [here](https://github.com/microsoft/qsharp-runtime/blob/df59efb7e11fc62439c777a85d0da95db4604d68/src/Simulation/Simulators/QCTraceSimulator/QCTraceSimulatorImpl.cs#L118), and I think it needs to be, since it can be modified.

samples/algorithms/integer-factorization/IntegerFactorization.csproj
```diff
@@ -5,4 +5,8 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.8.0" />
```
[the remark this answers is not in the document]
msoeken: Yes, that would be great to use.

samples/algorithms/integer-factorization/README.md
msoeken: It has to be captured from `STDOUT`. I will modify the README.

msoeken:
Thanks for the detailed comments, @cgranade. I addressed them.

cgranade (approved):
Looks good to me, thanks for adding this!

PR #471, opened 2021-03-09
Update lock-master.yml

transdigiware:


cgranade (changes requested):
Thanks for the contribution! Unfortunately, it looks like this change would break the workflow, as it is not valid YAML as written. I'm going to close the PR as such, but please feel free to reopen if you would like to update your PR.

PR #478, opened 2021-04-15
QIR Generation Testing

ScottCarda-MS:
Adds a check to the build script to generate QIR for several sample projects. As more projects become compatible, they can be added to the list, with the ultimate goal to have all the sample projects generate QIR, eventually. For now only some are compatible.


Build/build.ps1
```diff
@@ -36,55 +36,55 @@ Get-ChildItem (Join-Path $PSScriptRoot '..') -Recurse -Include '*.sln' `
     | ForEach-Object { Build-One $_.FullName $False }
 
 $QirProjects = @(
-    #(Join-Path $PSScriptRoot '..' 'samples' 'algorithms' 'chsh-game' 'CHSHGame.csproj'),
-    #(Join-Path $PSScriptRoot '..' 'samples' 'algorithms' 'database-search' 'DatabaseSearchSample.csproj'),
-    #(Join-Path $PSScriptRoot '..' 'samples' 'algorithms' 'integer-factorization' 'IntegerFactorization.csproj'),
-    (Join-Path $PSScriptRoot '..' 'samples' 'algorithms' 'oracle-synthesis' 'OracleSynthesis.csproj'),
```
cgranade: Out of curiosity, what was the motivation for this change? Is this needed to support deprecated versions of PowerShell (i.e.: 5.1)?

ScottCarda-MS: Join-Path doesn't support variadic arguments. If I wanted to pass arbitrarily many pieces of a path for it to glue together using the appropriate path separator for whatever system it is in, I need to use `[IO.Path]::Combine`.

cgranade: Join-Path has supported variadic arguments since 6.0 (7.1 is the current long-term release). Our build agents all use 6.0 and later, since 5.1 is deprecated and is not cross-platform.

ScottCarda-MS: @cgranade This is where I'm getting my information from: https://stackoverflow.com/a/25893843/5399520

swernli: Yeah, if you are running the build.ps1 scripts locally, you must do so in PowerShell 6.0+ aka pwsh (the product formerly known as PowerShell Core). You can install it from [here](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-windows?view=powershell-7.1). The PowerShell built into Windows is legacy and does not have the latest features, including security fixes.

As Chris points out, PowerShell 7.1 is the current release, and is what our build pipelines use. There `Join-Path` does support variadic arguments, and is what we use in many other build scripts, such as [here](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/qir-utils.ps1#L1) or [here](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/build-qir-runtime.ps1#L25).

ScottCarda-MS: The azp build was failing before, but perhaps it is because I was using quotes around the arguments? I'll try without the quotes. It was previously complaining about there being too many positional arguments to Join-Path, during the online build.

swernli: Oh! Then this is a bug in the `steps.yml`. Thanks for pointing that out! If you take a look at https://github.com/microsoft/Quantum/blob/main/Build/steps.yml#L29, you'll see that it is using the task name `powershell`, but it should be using `pwsh`. See [Using a specific shell](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell) in the GitHub Actions documentation. It turns out that `powershell` is the most confusing option of all of them, because on Windows it will use in-box PowerShell (the old version that doesn't support variadic arguments in `Join-Path`), but since that doesn't exist on other platforms, it will fall back to using PowerShell 7 everywhere else. This means now the scripts have to be written to the overlapping subset of functionality from legacy PowerShell and PowerShell 7, and is very annoying. We've updated the other repos to use `pwsh` in the yml instead, but I didn't realize samples repo here was still using the old way.

Can you update the steps.yml to use `pwsh` as part of your changes? It's the right fix to do, and it will also allow this script to start working as expected.

swernli: Here's the corresponding Azure DevOps documentation: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops

ScottCarda-MS: Cool that worked, thanks @swernli!

swernli (approved):
LGTM!

Build/build.ps1
```diff
@@ -8,15 +8,17 @@ $all_ok = $True
 
 function Build-One {
     param(
-        $project
+        $project,
+        $qirGenerationFlag
```
cgranade, line 12: Minor PowerShell style suggestion, but it may be more natural to make this a switch parameter instead:

```suggestion
        [switch]
        $GenerateQir
```

Doing so, line 21 would then become:

```
/property:QirGeneration=($GenerateQir ? "true" : "false")
```

Similarly, lines 90–91 become:

```
$QirProjects `
    | ForEach-Object { Build-One $_ -GenerateQir }
```

Build/build.ps1
```diff
@@ -31,7 +33,62 @@ function Build-One {
 }
 
 Get-ChildItem (Join-Path $PSScriptRoot '..') -Recurse -Include '*.sln' `
-    | ForEach-Object { Build-One $_.FullName }
+    | ForEach-Object { Build-One $_.FullName $False }
+
+$QirProjects = @(
+    #(Join-Path $PSScriptRoot .. samples algorithms chsh-game CHSHGame.csproj),
```
cgranade, line 39: I take it commented-out lines correspond to projects  not yet compatible with QIR? That seems like a reasonable exception to the general advice of not leaving commented-out code in the repo, but it may be helpful to have a quick comment explaining that.

cgranade (approved):
LGTM, thanks for doing this! Just had a couple minor comments.

Build/build.ps1
ScottCarda-MS: Yah good idea. Done.

Build/build.ps1
ScottCarda-MS: I'm not familiar with switch parameters, and I can't seem to get your suggestion to work properly. Could you please take another look at this?

cgranade: Sounds good, will do. It's also a minor comment at most; didn't mean to cause too much extra work, sorry.

Build/build.ps1
```diff
@@ -8,15 +8,18 @@ $all_ok = $True
...
     );
 
     Write-Host "##[info]Building $project..."
     dotnet build $project `
         -c $Env:BUILD_CONFIGURATION `
         -v $Env:BUILD_VERBOSITY `
         /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:ASSEMBLY_VERSION
+        /property:Version=$Env:ASSEMBLY_VERSION `
+        /property:QirGeneration=($generateQir ? "true" : "false")
```
cgranade, line 22: It looks like the subshell substitution here requires a `$`, sorry for the typo.
```suggestion
        /property:QirGeneration=$($generateQir ? "true" : "false")
```

ScottCarda-MS: Ah that did the trick. Thanks!

Build/build.ps1
ScottCarda-MS: Fixed

PR #482, opened 2021-04-29
Fix typo in Sudoku sample

cgranade:
As reported offline by @tcNickolas, a minor typo in a recent update to the Sudoku sample was not caught by our CI process as the sample project was incorrectly omitted from the relevant solution file. This PR:

- Fixes the typo.
- Adds the sudoku sample to the relevant sln file.
- Adds a new CI check to ensure all projects belong to at least one solution.

Note that this PR will **fail** as is, since there are other projects that fail the new check. Once CI has run and identified these projects, I will update the PR accordingly.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -27,6 +27,12 @@ jobs:
           | ForEach-Object {
               Write-Host "::error file=$_::Project file $_ should be in a solution."
             }
+
+        Write-Host "Found $($projectsInSolutions.Count) project files in solutions, and $($projectsNotInSolutions.Count) not correctly added to solutions.";
+        if ($projectsNotInSolutions.Count -gt 0) {
+          exit 1;
+        }
       name: "Ensure that all projects belong to a solution."
       shell: pwsh
+      continue-on-error: false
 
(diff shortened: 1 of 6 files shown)


cgranade:
It looks like the new CI check is working correctly, and is ensuring that all sample projects are actually included in CI builds. As such, this PR should be ready to go now.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,40 @@
...
+    - uses: actions/checkout@v2
+    - run: |
+        $solutions = Get-ChildItem -Recurse *.sln;
+        $projectsInSolutions = $solutions `
+          | ForEach-Object {
+            $sln = $_;
+            $slnRoot = [IO.Path]::GetDirectoryName($sln);
+            dotnet sln $sln list `
+              | Where-Object {
+                  $_.EndsWith(".csproj") -or $_.EndsWith(".fsproj")
```
tcNickolas, line 20: Are we willing to entertain the possibility of ever having .vsproj in our samples? :-)

tcNickolas (approved):

.github/workflows/checks.yml
cgranade: Happy to generalize this in the future, definitely. For now, just wanted to make sure to cover what project file types we already had in the repo.

PR #486, opened 2021-05-10
Add infrastructure to run samples as well as tests.

cgranade:
This PR adds infrastructure to CI to run Q# command line applications as well as running unit tests and checking builds. Not all projects are currently supported, but this should help improve our ability to catch breaking changes as early as possible.


msoeken (approved):
Looks good to me. Thanks @cgranade

anpaz (commented):
In principle I like the changes, however, do we know how much time this will be adding to the e2e build?
Samples are already the bottleneck: from a quick sampling of the latest builds they are consuming at least 50mins and up to 1:10hrs to run, making them the longest part of the build, so any time we add to them it's going to affect directly the overall time.
One way to mitigate this might be to take them out of the `build_main` Job so they don't block the start other validations. Let's take it offline if you would like to do this.

cgranade:
> In principle I like the changes, however, do we know how much time this will be adding to the e2e build?
> Samples are already the bottleneck: from a quick sampling of the latest builds they are consuming at least 50mins and up to 1:10hrs to run, making them the longest part of the build, so any time we add to them it's going to affect directly the overall time.

Fair enough; I made sure to denylist those samples that take longer to run, but I take the point. Happy to sync offline as you suggest. Let's include @bettinaheim as well, since this also affects our ability to test that QIR samples work for end-users.

cgranade:
> > In principle I like the changes, however, do we know how much time this will be adding to the e2e build?
> > Samples are already the bottleneck: from a quick sampling of the latest builds they are consuming at least 50mins and up to 1:10hrs to run, making them the longest part of the build, so any time we add to them it's going to affect directly the overall time.
> 
> Fair enough; I made sure to denylist those samples that take longer to run, but I take the point. Happy to sync offline as you suggest. Let's include @bettinaheim as well, since this also affects our ability to test that QIR samples work for end-users.

@anpaz, @bettinaheim: Following up on this quickly so as to unblock adding these validations, it looks like this PR adds about five to six minutes. That could possibly be reduced by parallelizing with `ForEach-Object -Parallel`, but that currently doesn't play well with the use of `$script:all_ok` to accumulate failed test steps.

anpaz (approved):
The latest e2e build shows the test.ps1 script taking about 15m, and the overall step for Quantum about 30m, which is about the same time it takes to build the runtime.
It's a bit too long, IMHO, but should be ok as right now it's not adding any significant time to the build.

cgranade:
> The latest e2e build shows the test.ps1 script taking about 15m, and the overall step for Quantum about 30m, which is about the same time it takes to build the runtime.
> It's a bit too long, IMHO, but should be ok as right now it's not adding any significant time to the build.

Sounds good, thank you for your review and your help addressing build time issues!

PR #494, opened 2021-05-18
Troubleshooting publishing issues, update metadata per guidance

Bradben:
Some samples aren't being published and indexed to the Samples Browser. These changes bring the metadata up to standard to test if that is the blocking issue. 

@cgranade, @KittyYeungQ    I can't assign reviewers in this repo, can you review and approve? Thanks

### DIFF
--- a/samples/algorithms/database-search/README.md
+++ b/samples/algorithms/database-search/README.md
@@ -7,19 +7,18 @@ languages:
 products:
 - qdk
 description: "This sample shows how to use Q# to search for a marked item with Grover's algorithm."
-extensions:
-  jupyter:
-    jupytext:
-      cell_markers: region,endregion
-      text_representation:
-        extension: .md
-        format_name: markdown
-        format_version: '1.2'
-        jupytext_version: 1.5.2
-    kernelspec:
-      display_name: .NET (PowerShell)
-      language: PowerShell
-      name: .net-powershell
+jupyter:
+  jupytext:
+    cell_markers: region,endregion
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.2'
+      jupytext_version: 1.5.2
+  kernelspec:
+    display_name: .NET (PowerShell)
+    language: PowerShell
+    name: .net-powershell
 ---
 
 # Database Search Sample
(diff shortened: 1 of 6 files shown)


samples/algorithms/database-search/README.md
```diff
@@ -7,18 +7,19 @@ languages:
 products:
 - qdk
 description: "This sample shows how to use Q# to search for a marked item with Grover's algorithm."
-jupyter:
```
cgranade, line 10: This will break mybinder.org deployment, but should not matter to the samples browser, as the workflow for stripping metadata removes this key: 

```
---
page_type: sample
languages:
- qsharp
- python
- csharp
products:
- qdk
description: "This sample shows how to use Q# to search for a marked item with Grover's algorithm."
---

# Database Search Sample
```

https://github.com/microsoft/Quantum/blob/%E2%AD%90samples/samples/algorithms/database-search/README.md

samples/azure-quantum/grover/README.md
```diff
@@ -1,14 +1,15 @@
 ---
 page_type: sample
-author: cgranade
```
cgranade, line 3: Elsewhere, we've been told that these keys are required; will they still work if they're scoped to `extensions`?

Bradben: They're required for topics that go to our usual OPS publishing, but not for the Samples Browser.  (Per the [sample metadata guidance](https://review.docs.microsoft.com/en-us/help/contribute/samples/process/onboarding?branch=master#supported-metadata-fields-for-readmemd) and poking around in many other Azure service samples).

My suspicion is that the Samples indexer doesn't like the metadata layout and isn't publishing the sample or indexing the azure-quantum service in the filter list. But I could be wrong.

samples/algorithms/database-search/README.md
Bradben: I suspected something like that.  There are 7 sample files that use the jupyter fields - 2 of them publish and 5 don't.  I'll have to run those by the samples team to see if they can spot something I'm not seeing.

samples/azure-quantum/grover/README.md
Bradben: @cgranade - can we push this through and see if it fixes the publishing for these samples?

PR #498, opened 2021-05-19
Add Azure Quantum chemistry sample notebooks

guenp:
Add the notebooks with the sample used in the //Build demo.


samples/azure-quantum/chemistry/data/broombridge/hydrogen_0.2.yaml
```diff
@@ -0,0 +1,56 @@
...
+        one_body_amplitudes:             # t^{q}_{p} p q 
+            - [0.001, "(2a)+", "(1a)"]
+            - [-0.001, "(2b)+", "(1b)"]
+        two_body_amplitudes: 			# t^{pq}_{rs} p q r s 			# If this is a PQQR term, the middle two indices must coincide.
+            - [-0.113, "(2a)+", "(2b)+", "(1a)", "(1b)"]
+  metadata: {molecule_name: H2}
+  n_electrons: 2
+  n_orbitals: 2
+  scf_energy: {units: hartree, value: -7.86099161}
+  scf_energy_offset: {units: hartree, value: 0.0}
```
cgranade, line 56:
```suggestion
  scf_energy_offset: {units: hartree, value: 0.0}

```

cgranade (commented):
I'm not done with my review yet, sorry (still need to review the notebooks), but wanted to share initial thoughts so as to unblock in the meantime. Thanks for opening this!

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
```diff
@@ -0,0 +1,43 @@
...
+    /// # Output
+    /// Tuple of estimated phase and energy.
+    operation GetEnergyQPE (
+        JWEncodedData: JordanWignerEncodingData,
+        nBitsPrecision : Int, 
+        trotterStepSize : Double, 
+        trotterOrder : Int
+    ) : (Double, Double) {
+
+        let (nSpinOrbitals, fermionTermData, inputState, energyOffset) = JWEncodedData!;
```
tcNickolas, line 37: You can replace a couple of unused variables:
```suggestion
        let (_, _, inputState, energyOffset) = JWEncodedData!;
```

samples/azure-quantum/chemistry/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Running the sample
+
+To run the sample, first make sure to install all [prerequisites](https://docs.microsoft.com/en-us/azure/quantum/install-python-qdk?tabs=tabid-conda) and start a Jupyter notebook with the following command:
+
+```cmd
+jupyter notebook
+```
+
+Open the notebooks ([QuantumPhaseEstimation/Molecule.ipynb](QuantumPhaseEstimation/Molecule.ipynb) or [Hamiltonian/Molecule.ipynb](Hamiltonian/Molecule.ipynb)) and execute the cells (shift + enter) to run the sample.
```
tcNickolas, line 27: Should we mention that the user needs AQ registration before running the second sample?

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "79c2f4e5",
+   "metadata": {},
+   "source": [
+    "# Approximate H2 molecule energy near ground state\n",
+    "\n",
+    "In this example notebook, we'll walk through calculating the energy of a molecule given an estimate for the ground state using Azure Quantum. We will do that using a simple algorithm where we prepare the quantum register in a state near the ground state of the molecule using Jordan-Wigner encoding, and evaluate the energy of the molecule by applying the Hamiltonian operator to that state. To be able to run the algorithm on near-term hardware, we will calculate the energy per Hamiltonian term, and then calculate the total energy by adding the outcome of these terms."
```
tcNickolas, line 10:
```suggestion
    "In this example notebook, we'll walk through calculating the energy of a molecule given an estimate for the ground state using Azure Quantum. We will do that using a simple algorithm where we prepare the quantum register in a state near the ground state of the molecule using Jordan-Wigner encoding, and evaluate the energy of the molecule by applying the Hamiltonian operator to that state. To be able to run the algorithm on near-term hardware, we will calculate the energy per Hamiltonian term, and then calculate the total energy by adding the values of these terms."
```

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9efb9be7",
+   "metadata": {},
+   "source": [
+    "### 1. Load H2 molecule data\n",
+    "\n",
+    "The pre-generated Broombridge file included in this sample contains the details on the molecule's Hamiltonian. A Hamiltonian is an operator that calculates the energy of a molecule by acting on a qubit register that represents the molecule's quantum state by Jordan-Wigner encoding. If we prepare the molecule in the ground state, then we know that applying the Hamiltonian we will calculate the ground state energy."
```
tcNickolas, line 42:
```suggestion
    "The pre-generated Broombridge file included in this sample contains the details on the molecule's Hamiltonian. A Hamiltonian is an operator that calculates the energy of a molecule by acting on a qubit register that represents the molecule's quantum state by Jordan-Wigner encoding. If we prepare the molecule in the ground state, then we know that applying the Hamiltonian will calculate the ground state energy."
```

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+      "text/plain": [
+       "Reloading workspace."
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "qsharp.reload()"
```
tcNickolas, line 124: For the demo, I combined all imports and reloads in one cell, since each of them is not particularly interesting on their own, just a prelude.

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "qsharp.azure.connect(\n",
+    "   resourceId=rid,\n",
+    "   location=\"West US\")"
```
tcNickolas, line 245: I think the location also depends on the user settings?

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   "source": [
+    "from Microsoft.Quantum.Samples.Chemistry.JordanWigner.Utils import ExpandedCoefficients"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3500a530",
+   "metadata": {},
+   "source": [
+    "Each Hamiltonian term has a term type, measurement operator and coefficient. To simplify this sample, we will move the coefficients into a flat list. Each nonzero coefficient will render a separate job that we send to Azure Quantum. To calculate the energy for that term, we map result from the job onto a -1 to 1 axis and multiply the value with the corresponding coefficient."
```
tcNickolas, line 355:
```suggestion
    "Each Hamiltonian term has a term type, measurement operator and coefficient. To simplify this sample, we will move the coefficients into a flat list. Each nonzero coefficient will render a separate job that we send to Azure Quantum. To calculate the energy for that term, we map result from the job onto a -1 to 1 axis and multiply the value by the corresponding coefficient."
```

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+       " {'0': 0.5, '1': 0.5}]"
+      ]
+     },
+     "execution_count": 20,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "results"
```
tcNickolas, line 519: Can combine this cell with the previous one, I generally try to avoid output-less cells

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+       " 'Succeeded']"
+      ]
+     },
+     "execution_count": 59,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "[qsharp.azure.status(j.id).status for j in jobs_hw]"
```
tcNickolas, line 716: For this cell, I added some more fields to make it a bit more visually interesting. In the demo I had

```
jobs_status = [qsharp.azure.status(j) for j in jobs_hw_past]
[j.status + " | " + j.name + " | " + j.target + " | " + j.creation_time + " | " + j.end_execution_time 
 for j in jobs_status]
```

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+     "execution_count": 100,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "calc_energy(coeffs, results)"
+   ]
+  },
+  {
```
tcNickolas, line 783: Do we need this last empty cell?

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,497 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Preparing Q# environment...\n",
+      "."
+     ]
+    }
+   ],
+   "source": [
+    "import qsharp"
```
tcNickolas, line 107: I'd combine those cells with imports and reloads

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,497 @@
...
+     "output_type": "stream",
+     "text": [
+      "CPU times: user 3.75 ms, sys: 4.01 ms, total: 7.76 ms\n",
+      "Wall time: 875 ms\n"
+     ]
+    },
+    {
+     "data": {
+      "text/plain": [
+       "(-0.2699806186678728, 882.6628565016248)"
```
tcNickolas, line 197: In the demo I modified the code so that it returned only one number, the energy. This was consistent with other method, and didn't introduce extra numbers that are not referred to later

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,497 @@
...
+   ],
+   "source": [
+    "GetEnergyQPE.estimate_resources(\n",
+    "    JWEncodedData=encoded_data_h2, \n",
+    "    nBitsPrecision=7,\n",
+    "    trotterStepSize=0.4,\n",
+    "    trotterOrder=1)"
+   ]
+  },
+  {
```
tcNickolas, line 468: I would add comparison with H2 VQE here, so that it's apples to apples

tcNickolas (commented):
I did a pass and left some comments. Thank you!

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// # Summary
+    /// Expands the compact representation of the Jordan-Wigner coefficients in order
+    /// to obtain a one-to-one mapping between these and Pauli terms.
+    ///
+    /// # Input
+    /// ## coeff
+    /// An array of coefficients, as read from the Jordan-Wigner Hamiltonian data structure.
+    /// ## termType
+    /// The type of the Jordan-Wigner term.
```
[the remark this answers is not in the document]
guenp: I'd love to add that but this function doesn't have that reference in the libraries either.

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
```diff
@@ -0,0 +1,127 @@
...
+        elif (termType == 3) {set nCoeffs = 8;}
+        else {set nCoeffs = 1;}
+
+        mutable coeffs = new Double[nCoeffs];
+
+        // Return the expanded array of coefficients
+        if ((termType == 0) or (termType == 1)) {
+            set coeffs w/= 0 <- coeff[0];
+        }
+        elif ((termType == 2) or (termType == 3)) {
```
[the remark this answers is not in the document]
guenp: This code was copied from the libraries (in order to be compatible with Azure Quantum) so I'd rather leave it like this to be consistent.

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
```diff
@@ -0,0 +1,127 @@
...
+
+        mutable ops = new Pauli[][nOps];
+
+        // Z and ZZ terms
+        if ((termType == 0) or (termType == 1)) {
+            mutable op = new Pauli[nQubits];
+            for idx in indices {
+                set op w/= idx <- PauliZ;
+            }
+            set ops w/= 0 <- op;
```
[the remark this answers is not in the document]
guenp: Thanks for the suggestion! I think we still need the line `set ops w/= 0 <- op;` though.

guenp:
@cgranade , thank you for your review! I'm seeing a lot of code style suggestions to the ChemUtils.qs file which is great; however, the contents of the file actually comes from the Q# libraries and changing the file here will introduce some discrepancies. Should we update this version and then also open a PR to the Q# Libraries, or what is your suggested approach here?

samples/azure-quantum/chemistry/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Manifest
+
```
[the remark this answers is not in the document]
guenp: Yes, thank you!

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
```diff
@@ -0,0 +1,43 @@
...
+        let (nQubits, (rescaleFactor, oracle)) = TrotterStepOracle(JWEncodedData, trotterStepSize, trotterOrder);
+        let statePrep = PrepareTrialState(inputState, _);
+        let phaseEstAlgorithm = RobustPhaseEstimation(nBitsPrecision, _, _);
```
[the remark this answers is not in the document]
guenp: I assumed those two are the same things (iterative PE is a type of QPE). Are they not?

samples/azure-quantum/chemistry/README.md
guenp: Added!

samples/azure-quantum/chemistry/data/broombridge/caffeine.yaml
```diff
@@ -0,0 +1,92 @@
+$schema: https://raw.githubusercontent.com/Microsoft/Quantum/master/Chemistry/Schema/broombridge-0.1.schema.json
```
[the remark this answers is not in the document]
cgranade: Likely, but by the same token as we need copyright headers on individual source files, it may be a good practice to make sure there's a header on these sources as well.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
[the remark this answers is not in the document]
cgranade: Iterative PE is an alternative to QPE in which the phase is read out classically (that is, not into a quantum register). Iterative PE methods can be much more efficient if you aren't using the phase as part of a quantum subroutine, such that reading out the phase classically suffices.

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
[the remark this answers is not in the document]
cgranade: As defined in the chemistry library, term types would not be passed directly by the user, but let me go grab a link.

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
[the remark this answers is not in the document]
cgranade: Ideally, we shouldn't be copying code from the libraries into samples — that's not a workflow that we would expect users to follow, such that the sample then is not representative of how to use Q# in practice. From that standpoint, if there's library code that needs to have a different or expanded API in order to support the limitations of current hardware, I think it'd make a lot of sense to open that as an API design proposal that we can support that use case for Q# library users more directly.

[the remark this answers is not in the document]
cgranade: With respect to code style, this is fairly old code that predates a lot of the more convenient features of Q#. While I don't think there's a great hurry to go rewrite library code that exists and works well as per unit test coverage (I've been doing so progressively as new features come along, but there's still a lot of code that hasn't been updated), I think the balance is again somewhat different for samples, in that we want the code to be read as an example of how to use Q# effectively to solve problems.

From that standpoint, making needed changes to the libraries can again reduce the volume of "old" code in the sample, and can help present a more idiomatic view of how to use Q#.

samples/azure-quantum/chemistry/data/broombridge/caffeine.yaml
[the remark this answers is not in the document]
guenp: I didn't see any copyright headers on the Broombridge files in the other samples either so I am not sure if we should add them here: https://github.com/microsoft/Quantum/tree/main/samples/chemistry/IntegralData/Broombridge_v0.2

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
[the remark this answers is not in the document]
guenp: It seems to me that they are similar enough to both be under the header "QPE", just "non-iterative" and "iterative". Agreed?

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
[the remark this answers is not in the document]
guenp: Sorry, perhaps I misunderstood your suggestion, what are you thinking I should add here?

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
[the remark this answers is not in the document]
guenp: I wasn't able to use the Q# libraries for this sample because it is incompatible with Azure Quantum, hence the code copying.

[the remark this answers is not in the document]
guenp: I've applied all your suggestions! The above one didn't work for me, I got a syntax error for this line:
```qsharp
return ConstantArray(nCoeffs, 0) w/ 0 <- coeff[0];
```

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   "source": [
+    "results"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f90d71ed",
+   "metadata": {},
+   "source": [
+    "### 5. Calculate energy"
```
cgranade, line 474:
```suggestion
    "### 5. Estimate energy"
```

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,1775 @@
...
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import qsharp\n",
+    "from qdk.chemistry import Molecule\n",
+    "from qdk.chemistry.broombridge import load_and_encode\n",
+    "qsharp.reload()\n",
```
cgranade, line 46: Similarly here, we should not need this line any longer to indicate dependencies from Q# workspaces on NuGet packages.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,1775 @@
...
+      "text/plain": [
+       "JsmolWidget(layout=Layout(align_self='stretch', height='400px'))"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "caffeine"
```
cgranade, line 90: It looks like this cell output isn't captured in nbviewer or GitHub's notebook renderer...?

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
```diff
@@ -0,0 +1,43 @@
...
+    /// ## nBitsPrecision
+    /// Number of bits of precision.
+    /// ## trotterStepSize
+    /// Trotter step size.
+    /// ## trotterOrder
+    /// The Trotter order to use.
+    ///
+    /// # Output
+    /// Tuple of estimated phase and energy.
+    operation GetEnergyQPE (
```
cgranade, line 30: Similarly here, I think "estimate" would be a more accurate verb to use; we in general don't have direct access to the energy, such that there is always finite uncertainty in our estimates.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,1775 @@
...
+    "resources_h2_vqe"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Comparing resources\n",
+    "\n",
+    "The below bar chart shows the comparison between number of Clifford gates for the different algorithms on both Caffeine and H2."
```
cgranade, line 504: If I understand right, this is per job submission (that is, max depth)? Seems really low for VQE over all to use 28 operations, since at the least you need the overhead for binomial accuracy (1% → ~ 10k shots).

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,1775 @@
...
+   "outputs": [],
+   "source": [
+    "prepare_state = qsharp.compile(\"\"\"open Microsoft.Quantum.Core;\n",
+    "open Microsoft.Quantum.Chemistry;\n",
+    "open Microsoft.Quantum.Chemistry.JordanWigner;\n",
+    "open Microsoft.Quantum.Chemistry.JordanWigner.VQE;\n",
+    "open Microsoft.Quantum.Intrinsic;\n",
+    "open Microsoft.Quantum.Diagnostics;\n",
+    "\n",
+    "operation PrepareStateVQE (JWEncodedData: JordanWignerEncodingData, theta1: Double, theta2: Double, theta3: Double) : Unit {\n",
```
cgranade, line 581: Out of curiosity, what motivated compiling this one inline, versus the other operations defined in the local workspace?

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
```diff
@@ -0,0 +1,1775 @@
...
+       "()"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "prepare_state.simulate(JWEncodedData=encoded_data_h2, theta1=0.001, theta2=-0.001, theta3=0.001, nSamples=1)"
```
cgranade, line 1148: This may be easier to read with `qsharp.config["dump.basisStateLabelingConvention"] = "bitstring"`. The default of `"littleEndian"` makes a lot of sense for arithmetic applications, but I think it'd be easier to see the structure of the single-excitation initial state as |1100⟩ than as |3⟩.

cgranade (commented):
Thanks for your responses and changes, @guenp! I took a more detailed read through, including the notebooks this time. Please let me know if there's anything I can expand on or help with. Thanks!

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   "source": [
+    "num_qubits, fermion_terms, _, energy_offset = encoded_data_h2"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "57757a37",
+   "metadata": {},
+   "source": [
+    "Reload the workspace to load the Q# program for import into Python."
```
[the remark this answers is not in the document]
guenp: Last time I checked we can't use `Microsoft.Quantum.Chemistry` because of the hardware limitations.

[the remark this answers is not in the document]
cgranade: I thought there were a few specific operations that couldn't be used? Regardless, `import qsharp.chemistry` causes the `Microsoft.Quantum.Chemistry` NuGet package to be loaded (hence why the `qsharp.reload()` is needed to inject that into the local workspace). Adding that package directly to the project file allows IQ# to know that you need Microsoft.Quantum.Chemistry from startup rather than compiling, failing, adding the package, and then reloading.

samples/azure-quantum/chemistry/Hamiltonian/ChemUtils.qs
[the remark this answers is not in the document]
guenp: I tried just using that function and I still got a compiler error. I think just including the chemistry library is already enough to trigger a failure.

samples/azure-quantum/chemistry/Hamiltonian/GetHamiltonianTerm.qs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// # Summary
+    /// Get the energy for a given measurement operator that corresponds
+    /// to a single term in the Hamiltonian.
+    ///
+    /// # Input
+    /// ## nOp
+    /// The operator to apply.
+    @EntryPoint()
+    operation GetHamiltonianTermH2 (nOp : Int) : Result {
```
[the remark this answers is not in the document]
guenp: What about `MeasureTermH2`?

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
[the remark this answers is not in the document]
guenp: I just did a quick web search and it looks like "iterative quantum phase estimation" is a term used in many publications.

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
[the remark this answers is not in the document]
guenp: I don't think we can include `Microsoft.Quantum.Chemistry` to a sample that is submitted to the Azure Quantum service. This is the reason why they are in two separate folders.

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   "source": [
+    "qsharp.reload()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2d96eb23",
+   "metadata": {},
+   "source": [
+    "The Q# program `GetHamiltonianTermH2` calculates the energy found by one of the Hamiltonian terms for H2 on a given ground state of the molecule.\n",
```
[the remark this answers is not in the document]
guenp: Sure, thanks for the suggestion!

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
```diff
@@ -0,0 +1,813 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "id": "15d7272c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from Microsoft.Quantum.Samples.Chemistry.JordanWigner.Utils import ExpandedCoefficients"
```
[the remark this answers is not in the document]
guenp: I'm not sure what you mean by "wrap in the local Q# workspace"?

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: I am not sure why but it doesn't work for me if I don't use `qsharp.reload()`.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: That's because it's an interactive JS widget.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/GetEnergyQPE.qs
guenp: I'd like to keep the names of the operations the same as in the notebook we used for the //build demo, if that's okay with you. I think it's clear from the context that we're estimating the energy, not getting the accurate value.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: The goal is not binomial accuracy, it's to see how many gates it requires per run. Adding more gates is difficult on near-term hardware but scaling the number of shots is trivial.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: This way the code is shown in the notebook.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: That's a helpful suggestion, thanks!

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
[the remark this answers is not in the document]
guenp: I've updated the description to use "estimate" rather than "calculate".

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
cgranade: Let me take a look, then, as it sounds like there may be a bug, then. We shouldn't need that here if there's a csproj that includes a package reference and `IQSharpLoadAutomatically`, so if that's not working we may need to open an IQ# issue.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
cgranade: Makes sense, but I think that may not have been clear from the notebook as written, such that clarifying that the resources count for VQE is measuring depth and not the total cost may be helpful here.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
cgranade: Makes sense; wasn't sure if there was another reason I was missing, so figured I'd ask. Thanks!

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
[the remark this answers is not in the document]
guenp: The reason why I added `qsharp.reload()` is that if there is a compilation error in the Q#, the import silently fails. `qsharp.reload()` triggers a compilation which then displays the error.

guenp:
@cgranade  thanks for your review! I believe I have addressed all your comments, some of them I wasn't able to comment on so I will do so here:

> As is, though, the name is inaccurate — this doesn't get a term. Even something like MeasureH2HamiltonianTerm may be more accurate. That doesn't strictly agree with the API principles definition for the verb Measure (doesn't take a register to measure), but it's at least somewhat closer?

I would like to keep the names the same as the notebook used in the MS Build demo if that's OK with you. I realize it's not 100% accurate but "Get" is a very generic word that doesn't necessarily imply accuracy. You could expand it with "Get the result of the measurement for the Hydrogen Hamiltonian term".

> I don't think it's technically correct to call iterative PE "QPE." At the least, it's not consistent with the definitions we've used in API names and in documentation so far.

Multiple publications refer to "iterative quantum phase estimation", see e.g. https://journals.aps.org/pra/abstract/10.1103/PhysRevA.76.030306. Is that literature referring to something else?

> Understood and agreed; that's why I want to find where we can improve the libraries to support users instead of setting the expectation that users need to copy source code from the libraries and modify it to make use of Azure Quantum platforms and services.

For some reason I cannot import the function directly from the chemistry library so I've added a wrapper function that uses the library version.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: Done, thanks!

samples/azure-quantum/chemistry/Hamiltonian/Approximate Energy using Jordan-Wigner transformation.ipynb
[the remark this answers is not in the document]
guenp: I think I understand what you meant here, I added a wrapper function `ExpandedCoefficients_`.

samples/azure-quantum/chemistry/QuantumPhaseEstimation/Molecule.ipynb
guenp: I figured out why it didn't work, there was a missing semicolon somewhere in the Q# file so the compilation silently failed such that importing via Python did not work. Adding `qsharp.reload()` here is therefore is just to trigger compilation and expose any errors in the Q# program (ie. not silently fail).

cgranade:
> @cgranade thanks for your review!

No worries, happy to help!

> I believe I have addressed all your comments, some of them I wasn't able to comment on so I will do so here:
> 
> > As is, though, the name is inaccurate — this doesn't get a term. Even something like MeasureH2HamiltonianTerm may be more accurate. That doesn't strictly agree with the API principles definition for the verb Measure (doesn't take a register to measure), but it's at least somewhat closer?
> 
> I would like to keep the names the same as the notebook used in the MS Build demo if that's OK with you. I realize it's not 100% accurate but "Get" is a very generic word that doesn't necessarily imply accuracy. You could expand it with "Get the result of the measurement for the Hydrogen Hamiltonian term".

That honestly raises a bit of an awkward tension. Ideally I'd have been able to review that in the //build/ demo itself, so that I could catch style issues that may lead to confusion. For inclusion in the samples, it is important to make sure that as much as possible we reduce confusion by following the conventions and notation documented in our style guide and API design principles, such that I'm reticent to forego our normal code review process for things like "Get" versus "Estimate" and "n" versus "idx." While these are admittedly minor issues, I do think it's still important that our samples are as consistent as possible.

At the same time, I definitely understand the need for consistency with our published video content as well.

> > I don't think it's technically correct to call iterative PE "QPE." At the least, it's not consistent with the definitions we've used in API names and in documentation so far.
> 
> Multiple publications refer to "iterative quantum phase estimation", see e.g. https://journals.aps.org/pra/abstract/10.1103/PhysRevA.76.030306. Is that literature referring to something else?

It's not entirely unheard of as terminology, but from my own expertise in phase estimation and from the consensus in literature, the term "quantum phase estimation" is most clearly understood to mean that the answer is read out into a quantum register for further coherent processing — that's not the case with iterative methods that read out to a classical inference procedure.

While that's not universally agreed upon, as you note, I think it is the most clear and unambiguous terminology that we can use in this case. This is the strategy, for example, behind the naming of Q# library operations for phase estimation, so adopting consensus terminology can help reduce confusion there as well.

Indeed, the original paper on robust phase estimation (https://arxiv.org/pdf/1502.02677.pdf) avoids the term "quantum phase estimation" in favor of less ambiguous terminology (there's one mention in an appendix when RPE is compared to known QPE bounds).

> > Understood and agreed; that's why I want to find where we can improve the libraries to support users instead of setting the expectation that users need to copy source code from the libraries and modify it to make use of Azure Quantum platforms and services.
> 
> For some reason I cannot import the function directly from the chemistry library so I've added a wrapper function that uses the library version.

Right, as discussed offline, host programs can only call into entry points. Q# libraries cannot expose entry points, so if you want to call directly into a Q# library function or operation from a host program, you need to "re-export" it in that way.

With respect to your point about `qsharp.reload()` being needed to raise compiler errors, that's an artifact of when we didn't have any way to load packages at the time of `import qsharp` statements. Now that we have csproj files supported in IQ#, hiding errors is no longer needed. I think I know of a good way we can do that, let me go on and file a feature request with that suggestion, then.

tcNickolas:
To clarify, consistency with the demo is explicitly not a goal for this sample. We cut a lot of material from the demo, including technical terminology such as QPE or VQE, both to fit in time and to make sure the content is accessible for developers who are not familiar with the terminology, and we don't have to do this in the sample.

cgranade:
IQ# issue is open at https://github.com/microsoft/iqsharp/issues/440.

guenp:
@tcNickolas that's fair, would you be OK if we link the build video in the README here, and make a note that the sample used in the video is an adapted version? I was hoping people would be able to use it as a tutorial.

guenp:
Thanks, @cgranade for opening that issue on IQ#!
As @tcNickolas we are OK diverging with names here. I was hoping to use the //build demo video as a tutorial, but it looks like a lot of the operation names are different anyway so we don't have to worry about being 100% consistent. If we do want to move closer to the //build video, we could rename `QPE` to `QuantumFCI`.

cgranade:
> Thanks, @cgranade for opening that issue on IQ#!

No worries, thanks for sharing your example here — helps understand where we can make improvements to IQ# going forward! ♥

guenp:
Thanks @cgranade ! I think for now I'd like to keep the naming consistent with the video and the Q# chemistry blog post where we refer to RPE as a category of quantum phase estimation. We can refactor later if any confusion arises.
Did you want to wait for that IQ# feature to be implemented or do you think this PR is ready to go?

guenp:
Tagging @dgwvfxn for review.

cgranade (approved):
As discussed in some comments here, and offline, I think we should still align this sample with the terminology and code style used elsewhere in the Quantum Development Kit (e.g. `n` vs `idx`, "iterative PE", and so forth). That said, I'm happy to open a new issue to track that and unblock for now, especially as there are similar issues elsewhere in docs and samples that would be good to address as part of that alignment.

Issue #496, opened 2021-05-19
QIR sample to automatically generate oracles from Boolean Q# functions


msoeken:
Implementing quantum oracles is difficult. Classical oracles are described as black boxes in description of algorithms (see, e.g., Hamiltonian simulation, Quantum phase estimation, or Grover’s algorithm). While the *quantum parts* (reflection operator in Grover, QFT in QPE) of the algorithm are described in detail, the oracle is a black box placeholder for some classical function, for which no implementation is provided, or at most for some example. I propose a QIR-based tool that can automatically generate Q# operations for such classical oracles from Q# functions.

In a first step the Q# functions should get Boolean inputs and outputs, e.g.:

```qsharp
function Majority3_Classical(a : Bool, b : Bool, c : Bool) : Bool {
  return (a or b) and (a or c) and (b or c);
}
```

If a corresponding intrinsic Q# operation is defined as follows:

```qsharp
operation Majority3(inputs : (Qubit, Qubit, Qubit), output : Qubit) : Unit {
  body intrinsic;
}
```

this tool automatically finds an implementation for `Majority3` based on a QIR representation of the program.

msoeken:
Closed by #499

PR #499, opened 2021-05-21
QIR oracle generation sample

msoeken:
Implements #496


samples/qir/oracle-generator/qsharp/Program.qs
```diff
@@ -0,0 +1,50 @@
...
+                        let result = IsResultOne(MResetZ(f));
+
+                        Message($"{cc} {cb} {ca} -> {result}");
+                    }
+                }
+            }
+        }
+    }
+
+    internal function InitOracleCompilerFor<'In, 'Out>(func : 'In -> 'Out) : Unit {
```
cgranade, line 38: What is this function used for?

samples/qir/oracle-generator/qsharp/Program.qs
```diff
@@ -0,0 +1,50 @@
...
+        let _ = Microsoft.Quantum.Intrinsic.X;
+        let _ = Microsoft.Quantum.Intrinsic.CNOT;
+        let _ = Microsoft.Quantum.Intrinsic.CCNOT;
+        let _ = func;
+    }
+}
+
+namespace OracleCompiler.Classical {
+    internal function Majority3(a : Bool, b : Bool, c : Bool) : Bool {
```
cgranade, line 47: Some comments here could help as well.

samples/qir/oracle-generator/qsharp/project.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.Quantum.Sdk/0.16.2105142209-alpha">
```
cgranade, line 1: Does this require an alpha version? Is it possible to use a released version?

cgranade (commented):
This review is preliminary, given the size of your PR, but wanted to give early feedback. Thanks!

cgranade:
It looks like the build failure is because the agent tests against .NET Core SDK 3.1 only, matching our end-user installation instructions. We should discuss how to generalize the build for this repo since this sample requires 5.0.

samples/qir/oracle-generator/oracle-compiler/oracle_compiler.cpp
```diff
@@ -0,0 +1,144 @@
+#include <fstream>
+#include <iostream>
+#include <unordered_map>
+
+#include <fmt/format.h>
```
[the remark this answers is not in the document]
msoeken: I totally agree. I will spend some time to add documentation to all these files.

samples/qir/oracle-generator/oracle-compiler/oracle_compiler.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    future::xag_minmc_resynthesis<xag_network> resyn;
+
+    auto mc = *multiplicative_complexity(xag);
+    fmt::print("[i] initial XAG from LLVM: {} AND gates, {} XOR gates\n", mc, xag.num_gates() - mc);
+
+    /* collapse and resynthesize if number of inputs is small */
+    if (xag.num_pis() <= 8)
+    {
+        auto klut = single_node_network<klut_network>(xag);
```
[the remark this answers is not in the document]
msoeken: Exactly, I'll also add comments here.

samples/qir/oracle-generator/qsharp/Program.qs
msoeken: I'll add a comment. But for now it's actually only used as a workaround to avoid that all the operations and functions referred to in this function are included into the generated QIR code. If they are not used, they are going to be stripped.

samples/qir/oracle-generator/qsharp/project.csproj
msoeken: I am not sure, I'll try to use the version of the upcoming release.

msoeken:
> It looks like the build failure is because the agent tests against .NET Core SDK 3.1 only, matching our end-user installation instructions. We should discuss how to generalize the build for this repo since this sample requires 5.0.

I am not even sure whether .NET 5 is required. I am using it as a default when setting up projects. I'll check whether it can work with 3.1 to avoid this error.

samples/qir/oracle-generator/qsharp/Program.qs
cgranade: Makes sense, I figured it was something like that. As you say, adding a comment can help the reader understand why that workaround is needed.

cgranade:
> > It looks like the build failure is because the agent tests against .NET Core SDK 3.1 only, matching our end-user installation instructions. We should discuss how to generalize the build for this repo since this sample requires 5.0.
> 
> I am not even sure whether .NET 5 is required. I am using it as a default when setting up projects. I'll check whether it can work with 3.1 to avoid this error.

Makes sense. At the moment, the Quantum Development Kit officially requires 3.1 to be installed, since that's the most up-to-date LTS release, such that if our samples can work for users following those setup instructions that would be ideal. Thanks for looking into this!

samples/qir/oracle-generator/qsharp/project.csproj
msoeken: The package `Microsoft.Quantum.Qir.Runtime` does not exist in a released version yet. I suggest to wait until the May release and then change the version number, and keep this alpha version for now.

msoeken:
Thanks a lot for your comments @cgranade. I addressed all but one of your comments and also added more comments to the source files. The missing comment of yours concerns whether we can use a released version for this sample, but none of the released version suffice so far. Therefore, I suggest to address this comment once the next release is out.

samples/qir/oracle-generator/qsharp/project.csproj
cgranade: Makes sense; just wanted to double-check on that. I think it's fine to keep the version number as-is, then, but I would emphasize in the README that QIR generation is still in preview to help set user expectations.

msoeken:
@cgranade Thanks for all your comments. I have just recently updated the README to meet the expectations of the sample. Have all your comments been addressed and can the PR be approved?

samples/interoperability/qrng/NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="qdk-alpha" value="https://pkgs.dev.azure.com/ms-quantum-public/Microsoft Quantum (public)/_packaging/alpha/nuget/v3/index.json" protocolVersion="3" />
+    <add key="local" value="/home/mathias/code/nugets" />
```
cgranade, line 5: It looks like this line may not have been intended for inclusion in the sample?

samples/qir/oracle-generator/oracle-generator/oracle_generator.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <mockturtle/algorithms/gates_to_nodes.hpp>
+#include <mockturtle/algorithms/node_resynthesis.hpp>
+#include <mockturtle/algorithms/node_resynthesis/shannon.hpp>
+#include <mockturtle/algorithms/node_resynthesis/xag_minmc2.hpp>
+#include <mockturtle/networks/abstract_xag.hpp>
+#include <mockturtle/networks/klut.hpp>
+#include <mockturtle/networks/xag.hpp>
+#include <mockturtle/utils/cost_functions.hpp>
+
+// W modify LLVM through the object model provided by the LLVM C++ API.
```
cgranade, line 23:
```suggestion
// We modify LLVM through the object model provided by the LLVM C++ API.
```

cgranade (approved):
LGTM! Just had a couple minor comments left.

samples/interoperability/qrng/NuGet.config
msoeken: In fact, the whole file is not intended for inclusion. Might have added it by mistake.

PR #503, opened 2021-05-28
Add IonQ cost to Azure Quantum samples

dime10:
@KittyYeungQ

Teleportation sample is exclude since it's not compatible with IonQ.

Also update certain sample inputs as follows:
 - grover: `--n-qubits` 4 doesn't run on Honeywell, change to 3 (too many qubits, 7 in total)
 - hidden-shift: `--n-qubits` avoid confusion in regards to which input is used in the cost estimate, simulator (4) or AQ (3), make both 3

### DIFF
--- a/samples/azure-quantum/grover/README.md
+++ b/samples/azure-quantum/grover/README.md
@@ -41,7 +41,7 @@ az quantum target list --output table
 ```
 
 > :warning:
-> This sample makes use of paid services on Azure Quantum. The cost of running this sample *with the provided parameters* on IonQ in a Pay-As-You-Go subscription is approximately $14 USD (or the equivalent amount in your local currency). This quantity is only an approximate estimate and should not be used as a binding reference. The cost of the service might vary depending on your region, demand and other factors.
+> This sample makes use of paid services on Azure Quantum. The cost of running this sample *with the provided parameters* on IonQ in a Pay-As-You-Go subscription is approximately $5 USD (or the equivalent amount in your local currency). This quantity is only an approximate estimate and should not be used as a binding reference. The cost of the service might vary depending on your region, demand and other factors.
 
 ## Manifest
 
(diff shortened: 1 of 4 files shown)


samples/azure-quantum/grover/README.md
```diff
@@ -40,6 +40,9 @@ For a full list of available QIO and quantum computing targets, run:
 az quantum target list --output table
 ```
 
+> ::warning::
```
cgranade, line 43: This way of formatting warnings is specific to docs.microsoft.com, and may not look great as rendered by GitHub's web UI or by mybinder.org. I'd suggest double-checking both of those renderings if possible.

cgranade (commented):
Looks good at a first glance, but I had two quick questions:

- How do we keep this up to date as pricing changes, or as we maintain and develop these samples?
- Does it make sense to also give estimates for other providers to avoid the impression that these samples are IonQ-specific?

samples/azure-quantum/grover/README.md
dime10: I thought the docs specific way would be `> [!WARNING]`.
I saw this syntax in a repo readme, which renders a small icon in the GitHub interface. Looking at GFM it turns out they are just emoji codes. I made a typo though in using double colons, should be single colons.

dime10:
Thanks @cgranade, these are good points.

1. The idea is to add rough order of magnitude estimates so the user knows what they are getting into when running the sample. This has been a popular request. But it is true that it is susceptible to price changes. @KittyYeungQ 
2. The plan is to add Honeywell costs very shortly as soon as I have collected data on it.

dime10:
Since Honeywell discontinued H0, this PR will not be updated with pay-as-you-go cost estimates for Honeywell.

cgranade (approved):

PR #513, opened 2021-06-16
QIR Validation Script

ScottCarda-MS:
Moving the QIR validation out of the build script and into its own separate script. This script may be run by itself manually, but its ultimate purpose is to be run as part of the QDK's CI validation steps.

Several Samples are currently in a good state to be compiled with QIR, have the resulting dll compiled further into an executable, and have that executable run to test that the QIR is working as intended. Many samples have some qualities about them that make them not compatible for this process as of yet. The purpose of this PR is not to make all the samples compatible for QIR validation but to set up a framework for QIR validation using some of the already-compatible samples.

### DIFF
--- a/Build/QIR_validation.ps1
+++ b/Build/QIR_validation.ps1
@@ -1,0 +1,118 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+$ErrorActionPreference = 'Stop'
+
+& "$PSScriptRoot/set-env.ps1"
+$all_ok = $True
+
+function Build-One {
+    param(
+        $project
+    );
+
+    $project_directory = $(Split-Path $project -Parent)
+    $project_name = $(Split-Path $project -LeafBase)
+
+    pushd $project_directory
+
+    Write-Host "##[info]Building $project..."
+    dotnet build $project `
+        -c $Env:BUILD_CONFIGURATION `
+        -v $Env:BUILD_VERBOSITY `
+        --no-dependencies `
+        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
+        /property:Version=$Env:ASSEMBLY_VERSION `
+        /property:QirGeneration="true"
+
+    if  ($LastExitCode -ne 0) {
+        Write-Host "##vso[task.logissue type=error;]Failed to build $project."
+        $script:all_ok = $False
+    } else {
+        qir-cli `
+            --dll (Join-Path bin $Env:BUILD_CONFIGURATION netcoreapp3.1 "${project_name}.dll") `
+            --exe qir
+        if  ($LastExitCode -ne 0) {
+            Write-Host "##vso[task.logissue type=error;]Failed to build $project."
+            $script:all_ok = $False
+        } else {
+            Get-ChildItem (Join-Path qir *__Interop.exe) `
+                | ForEach-Object { & $_ }
+            if  ($LastExitCode -ne 0) {
+                Write-Host "##vso[task.logissue type=error;]$project encountered an error or failed during execution."
+                $script:all_ok = $False
+            }
+        }
+    }
+
+    if ($env:FORCE_CLEANUP -eq "true") {
+        # Force cleanup of generated bin, obj, and qir folders for this project.
+        Write-Host "##[info]Cleaning up bin/obj from $project_directory..."
+        Get-ChildItem -Path $project_directory -Recurse | Where-Object { ($_.name -eq "bin" -or $_.name -eq "obj" -or $_.name -eq "qir") -and $_.attributes -eq "Directory" } | Remove-Item -recurse -force
+    }
+
+    popd
+}
+
+# The commented out lines are sample projects that are not yet compatible for QIR generation/execution. 
+# 'not compatible' means that the structure of the sample is not compatible for QIR generation.
+# 'needs argument(s)' means that the sample can generated QIR, but running the exe requires command line arguments.
+# 'package version issue' means there is an issue with the versioning of the packages involved, disallowing QIR generation.
+$QirProjects = @(
+    # not compatible #(Join-Path $PSScriptRoot .. samples algorithms chsh-game CHSHGame.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples algorithms database-search DatabaseSearchSample.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples algorithms integer-factorization IntegerFactorization.csproj),
+    (Join-Path $PSScriptRoot .. samples algorithms oracle-synthesis OracleSynthesis.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples algorithms order-finding OrderFinding.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples algorithms repeat-until-success RepeatUntilSuccess.csproj),
+    (Join-Path $PSScriptRoot .. samples algorithms reversible-logic-synthesis ReversibleLogicSynthesis.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples algorithms simple-grover SimpleGroverSample.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples algorithms sudoku-grover SudokuGroverSample.csproj),
+
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples azure-quantum grover Grover.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples azure-quantum hidden-shift HiddenShift.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples azure-quantum ising-model IsingModel.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples azure-quantum parallel-qrng ParallelQrng.csproj),
+    # package version issue #(Join-Path $PSScriptRoot .. samples azure-quantum teleport Teleport.csproj),
+
+    (Join-Path $PSScriptRoot .. samples characterization phase-estimation PhaseEstimationSample.csproj),
+
+    # not compatible #(Join-Path $PSScriptRoot .. samples chemistry AnalyzeHamiltonian 1-AnalyzeHamiltonian.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples chemistry CreateHubbardHamiltonian CreateHubbardHamiltonian.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples chemistry GetGateCount 3-GetGateCount.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples chemistry MolecularHydrogen MolecularHydrogen.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples chemistry RunSimulation 2-RunSimulation.csproj),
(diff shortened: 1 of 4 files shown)


Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+        Write-Host "##[info]Cleaning up bin/obj from $project_directory..."
+        Get-ChildItem -Path $project_directory -Recurse | Where-Object { ($_.name -eq "bin" -or $_.name -eq "obj" -or $_.name -eq "qir") -and $_.attributes -eq "Directory" } | Remove-Item -recurse -force
+    }
+
+    popd
+}
+
+# The commented out lines are sample projects that are not yet compatible for QIR generation/execution. 
+# 'not compatible' means that the structure of the sample is not compatible for QIR generation.
+# 'needs argument(s)' means that the sample can generated QIR, but running the exe requires command line arguments.
```
ScottCarda-MS, line 59: There are quite a lot of samples that can be compiled with QIR, but not executed. I had wanted to find a way to add some switch logic to the script that allowed me to specify per sample how far I wanted to go with the validation: build dll, build executables, run executables. Depending on the current blockers for a given sample, it may not be able to be run, but it can still be useful to compile to tests that part of the QIR compilation process, for example. Have a switch like this would also make it easier to onboard samples that are currently not compatible as they could be onboarded in stages.

I don't have much experience with powershell, so I wasn't able to figure out how to implement this switch logic, but maybe someone with more experience can show me how something like that would look.

ScottCarda-MS: It would be great to figure out how to run samples that need arguments in a programmatic way. I guess the alternative would be to hardcode the needed arguments for each sample that needs them. If that is the direction to go it, I would need some help with the powershell in learning how to do that.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+    if ($env:FORCE_CLEANUP -eq "true") {
+        # Force cleanup of generated bin, obj, and qir folders for this project.
+        Write-Host "##[info]Cleaning up bin/obj from $project_directory..."
+        Get-ChildItem -Path $project_directory -Recurse | Where-Object { ($_.name -eq "bin" -or $_.name -eq "obj" -or $_.name -eq "qir") -and $_.attributes -eq "Directory" } | Remove-Item -recurse -force
+    }
+
+    popd
+}
+
+# The commented out lines are sample projects that are not yet compatible for QIR generation/execution.
```
ScottCarda-MS, line 57: It would be great to figure out how to run samples that need arguments in a programmatic way. I guess the alternative would be to hardcode the needed arguments for each sample that needs them. If that is the direction to go it, I would need some help with the powershell in learning how to do that.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+    (Join-Path $PSScriptRoot .. samples getting-started measurement Measurement.csproj),
+    (Join-Path $PSScriptRoot .. samples getting-started qrng Qrng.csproj),
+    # needs argument(s) #(Join-Path $PSScriptRoot .. samples getting-started simple-algorithms SimpleAlgorithms.csproj),
+    (Join-Path $PSScriptRoot .. samples getting-started teleportation TeleportationSample.csproj),
+
+    # not compatible #(Join-Path $PSScriptRoot .. samples machine-learning half-moons HalfMoons.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples machine-learning parallel-half-moons ParallelHalfMoons.csproj),
+    # not compatible #(Join-Path $PSScriptRoot .. samples machine-learning wine Wine.csproj),
+
+    # package version issue #(Join-Path $PSScriptRoot .. samples numerics CustomModAdd CustomModAdd.csproj),
```
bettinaheim, line 99: What is the package version issue?

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+$ErrorActionPreference = 'Stop'
+
+& "$PSScriptRoot/set-env.ps1"
+$all_ok = $True
+
+function Build-One {
```
bettinaheim, line 9: I'd suggest to take both the project as well as the command line arguments here for execution.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+        $script:all_ok = $False
+    } else {
+        qir-cli `
+            --dll (Join-Path bin $Env:BUILD_CONFIGURATION netcoreapp3.1 "${project_name}.dll") `
+            --exe qir
+        if  ($LastExitCode -ne 0) {
+            Write-Host "##vso[task.logissue type=error;]Failed to build $project."
+            $script:all_ok = $False
+        } else {
+            Get-ChildItem (Join-Path qir *__Interop.exe) `
```
bettinaheim, line 39: Can't you just pass the command line arguments here?

bettinaheim (approved):
I am good with taking it in as is, so that we can first set up the build to run these and then work on adding support for samples with command line arguments etc.

Build/QIR_validation.ps1
bettinaheim: If by hardcoding you mean hardcoding the values in this powershell script, then I think that makes sense.

Build/QIR_validation.ps1
bettinaheim: Let's focus on having everything that builds also be able to run - it seems that's more of a setup issue and it should never be the case that something that builds can't run.

Build/QIR_validation.ps1
ScottCarda-MS: I'm not sure how. I don't know if there is any way to automatically discover command line arguments for an given executable.

Build/QIR_validation.ps1
ScottCarda-MS: I'm looking for fairly detailed help with this, as I am not very skilled with powershell.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) Microsoft Corporation.
```
cgranade, line 1: Minor, but our other build scripts have generally had all-lowercase names separated by dashes with the relevant verb first (e.g.: `validate-qir.ps1` or `test-qir.ps1`).

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+# Licensed under the MIT License.
+
+$ErrorActionPreference = 'Stop'
+
+& "$PSScriptRoot/set-env.ps1"
+$all_ok = $True
+
+function Build-One {
+    param(
+        $project
```
cgranade, line 11: Two PowerShell style nits:

- Parameter names should be in `$PascalCase`.
- Variable names should be in `$camelCase`.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+    );
+
+    $project_directory = $(Split-Path $project -Parent)
+    $project_name = $(Split-Path $project -LeafBase)
+
+    pushd $project_directory
```
cgranade, line 17: `pushd` may fail on some macOS and Linux systems, as they may have binaries with the same name that override the shell definition of `pushd`. In general, `Push-Location` and `Pop-Location` are more portable cross-platform.

Build/QIR_validation.ps1
```diff
@@ -0,0 +1,118 @@
...
+        --no-dependencies `
+        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
+        /property:Version=$Env:ASSEMBLY_VERSION `
+        /property:QirGeneration="true"
+
+    if  ($LastExitCode -ne 0) {
+        Write-Host "##vso[task.logissue type=error;]Failed to build $project."
+        $script:all_ok = $False
+    } else {
+        qir-cli `
```
cgranade, line 32: Is it required that this tool is already installed? If so, checking for the command earlier and providing an actionable error message when it is missing can help both internal and external developers working with this script. To check if a command is missing:

```powershell
if (-not (Get-Command qir-cli -ErrorAction SilentlyContinue)) {
    Write-Host "##[error]The qir-cli command is missing; you can install it by running `TODO`.";
}
```

Build/QIR_validation.ps1
cgranade: One way to do that is to store a PowerShell array of needed arguments with each project listed here, then to splat those arguments into the call above.

https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_splatting

Build/QIR_validation.ps1
ScottCarda-MS: done

ScottCarda-MS: changed to `validate-qir.ps1`

Build/QIR_validation.ps1
ScottCarda-MS: done

Build/QIR_validation.ps1
ScottCarda-MS: changed to Push-Location and Pop-Location

Build/QIR_validation.ps1
ScottCarda-MS: Yah this is a really good idea. I've added a check like this.

Build/QIR_validation.ps1
ScottCarda-MS: I'll try to be more specific here. What I am looking for is some way to zip the arguments and their values together with the path to the csproj file in the big list of samples. So instead of having
```
$QirProjects = {
    <path to sample 1>,
    <path to sample 2>,
    ...
}
```
I would have something like
```
$QirProjects = {
    { Path = <path to sample 1>, Args = [(arg1,val1), (arg2,val2), ...] },
    { Path = <path to sample 2>, Args = [(arg1,val1), (arg2,val2), ...] },
    ...
}
```
but I'm not seeing a straightforward way of accomplishing that. I expect that something like this is built in to powershell, or that maybe I'm just not thinking about the problem in a powershell-like way.

ScottCarda-MS: I guess that might be further complicated by the fact that any given sample may have multiple entry points, resulting in multiple executables to be called, possibly with different arguments.

cgranade: Setting aside the multiple entry points part, your snippet above is almost what you need:

```powershell
$qirProjects = @(
    @{ "Path" = "path to sample 1"; "Args" = @("--arg-1", "val1", "--arg-2", "val2"); }
    ...
);
```

You can then splat in `$_["Args"]` in your `ForEach-Object` script blocks, using the splatting syntax linked to above.

ScottCarda-MS: Oh a hash table, yah that's perfect, exactly what I need. Thanks @cgranade! I can see now what is meant by 'splat'. It is similar to the star syntax from python or 'explode' from other languages.

Build/QIR_validation.ps1
ScottCarda-MS: done

Build/QIR_validation.ps1
ScottCarda-MS: done

Build/QIR_validation.ps1
ScottCarda-MS: sounds good to me

Build/QIR_validation.ps1
ScottCarda-MS: I went back to recreate the issues to give you some specifics to answer your question, and found that I couldn't recreate the issues. I've removed those designations and included those samples. The package version might just be something to keep an eye on as this becomes a validation step in the CI.

PR #519, opened 2021-07-01
QIR Validation Script - Quantum

ScottCarda-MS:
Major updates to the validation script.

See also: https://github.com/microsoft/qsharp-runtime/pull/763


Build/validate-qir.ps1
```diff
@@ -30,19 +30,33 @@ function Build-One {
...
-            Write-Host "##vso[task.logissue type=error;]Failed to build $Project."
+        if  ($LastExitCode -ne 0 -or (Get-ChildItem (Join-Path qir *__Interop.exe)).count -eq 0) {
+            Write-Host "##vso[task.logissue type=error;]Failed to compile $projectName to executable."
             $script:allOk = $False
         } else {
+            Write-Host "##[info]Updating environment path variables."
+            $qirPath = Join-Path $projectDirectory qir
+            $old_DYLD_LIBRARY_PATH = $env:DYLD_LIBRARY_PATH
+            $old_LD_LIBRARY_PATH = $env:LD_LIBRARY_PATH
+            $env:DYLD_LIBRARY_PATH += ":" + $qirPath + ":"
```
cgranade, line 45:
```suggestion
            $Env:DYLD_LIBRARY_PATH += ":${qirPath}:"
```
Note that the `{}` are required here because of the colons following the name of the variable. Without `{}`, `qirPath` is interpreted as a variable scope similar to `Env`, `Script`, `Global`, or `Using`:
image

Build/validate-qir.ps1
```diff
@@ -30,19 +30,33 @@ function Build-One {
...
             --exe qir
-        if  ($LastExitCode -ne 0) {
-            Write-Host "##vso[task.logissue type=error;]Failed to build $Project."
+        if  ($LastExitCode -ne 0 -or (Get-ChildItem (Join-Path qir *__Interop.exe)).count -eq 0) {
+            Write-Host "##vso[task.logissue type=error;]Failed to compile $projectName to executable."
             $script:allOk = $False
         } else {
+            Write-Host "##[info]Updating environment path variables."
+            $qirPath = Join-Path $projectDirectory qir
+            $old_DYLD_LIBRARY_PATH = $env:DYLD_LIBRARY_PATH
```
cgranade, line 43: This looks like it sets library paths for macOS and Linux hosts, is anything needed for Windows hosts?

Build/validate-qir.ps1
```diff
@@ -30,19 +30,33 @@ function Build-One {
         Write-Host "##vso[task.logissue type=error;]Failed to build $Project."
         $script:allOk = $False
     } else {
+        Write-Host "##[info]Compiling $projectName to executable..."
         qir-cli `
             --dll (Join-Path bin $Env:BUILD_CONFIGURATION netcoreapp3.1 "${projectName}.dll") `
             --exe qir
-        if  ($LastExitCode -ne 0) {
-            Write-Host "##vso[task.logissue type=error;]Failed to build $Project."
+        if  ($LastExitCode -ne 0 -or (Get-ChildItem (Join-Path qir *__Interop.exe)).count -eq 0) {
```
cgranade, line 37:
```suggestion
        if  ($LastExitCode -ne 0 -or (Get-ChildItem (Join-Path qir *__Interop.exe) -ErrorAction SilentlyContinue).Count -eq 0) {
```

Build/validate-qir.ps1
```diff
@@ -57,23 +71,24 @@ function Build-One {
...
 $qirProjects = @(
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples algorithms chsh-game CHSHGame.csproj); Args = @() },
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples algorithms database-search DatabaseSearchSample.csproj); Args = @() },
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples algorithms integer-factorization IntegerFactorization.csproj); Args = @() },
     @{ Path = (Join-Path $PSScriptRoot .. samples algorithms oracle-synthesis OracleSynthesis.csproj); Args = @() },
     @{ Path = (Join-Path $PSScriptRoot .. samples algorithms order-finding OrderFinding.csproj); Args = @("--index", "1") },
-    @{ Path = (Join-Path $PSScriptRoot .. samples algorithms repeat-until-success RepeatUntilSuccess.csproj); Args = @("--gate", "simple", "--input-value", "true", "--input-basis", "PauliX", "--limit", "4", "--num-runs", "2") },
+    @{ Path = (Join-Path $PSScriptRoot .. samples algorithms repeat-until-success RepeatUntilSuccess.csproj); Args = @("--gate", "simple", "--inputValue", "true", "--inputBasis", "PauliX", "--limit", "4", "--numRuns", "2") },
```
cgranade, line 82: Perhaps tangential, but why are the input parameter names different here than what's generated for stand-alone executables?

Build/validate-qir.ps1
```diff
@@ -57,23 +71,24 @@ function Build-One {
 
 # The commented out lines are sample projects that are not yet compatible for QIR generation/execution. 
 # 'not compatible' means that the structure of the sample is not compatible for QIR generation.
-# 'needs argument(s)' means that the sample can generated QIR, but running the exe requires command line arguments.
+# 'relies on hardware' means that the sample targets a hardware architecture that is not yet supported for QIR.
+# 'memory leak' means that the sample stalls when running and uses way too many system resources.
```
cgranade, line 75: Is there a bug filed that tracks these memory leaks? Would be good to link in comments so that there's an actionable path to adding affected samples back into validation.

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
-    # needs argument(s) #@{ Path = (Join-Path $PSScriptRoot .. samples simulation qaoa QAOA.csproj); Args = @() }
+    @{ Path = (Join-Path $PSScriptRoot .. samples simulation qaoa QAOA.csproj); Args = @("--numTrials", "20") }
 )
 
+Write-Host "##[info]Beginning Microsoft.Quantum.Qir.CommandLineTool installation."
 if (-not (Get-Command qir-cli -ErrorAction SilentlyContinue)) {
-    Write-Host "##[error]The qir-cli command is missing; you can install it by running `dotnet tool install Microsoft.Quantum.Qir.CommandLineTool -g`.";
-    $script:allOk = $False
+    $version = ($Env:NUGET_VERSION -split "-")[0] + "-alpha"
+    Write-Host "##[info]Going to install Microsoft.Quantum.Qir.CommandLineTool.$version."
```
cgranade, line 131: It feels very surprising for the validation script to make a global change to the installation environment. This also duplicates a lot of functionality with that provided by tool manifests... I know that approach had been discussed and rejected earlier, but it may be worth revisiting if it can help avoid making global changes from within an integration testing script.

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples simulation ising generators IsingGeneratorsSample.csproj); Args = @() },
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples simulation ising phase-estimation IsingPhaseEstimationSample.csproj); Args = @() },
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples simulation ising simple SimpleIsingSample.csproj); Args = @() },
     # not compatible #@{ Path = (Join-Path $PSScriptRoot .. samples simulation ising trotter-evolution IsingTrotterSample.csproj); Args = @() },
-    # needs argument(s) #@{ Path = (Join-Path $PSScriptRoot .. samples simulation qaoa QAOA.csproj); Args = @() }
+    @{ Path = (Join-Path $PSScriptRoot .. samples simulation qaoa QAOA.csproj); Args = @("--numTrials", "20") }
 )
 
+Write-Host "##[info]Beginning Microsoft.Quantum.Qir.CommandLineTool installation."
 if (-not (Get-Command qir-cli -ErrorAction SilentlyContinue)) {
```
cgranade, line 129: Is it important to validate that the right version of `qir-cli` is installed?

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
+        $script:allOk = $False
+    } else {
+        Write-Host "##[info]The Microsoft.Quantum.Qir.CommandLineTool.$version has been installed."
+    }
 } else {
+    Write-Host "##[info]The Microsoft.Quantum.Qir.CommandLineTool is already installed."
+}
+
+if ($allOk) {
+    if ($IsWindows) {
```
cgranade, line 144: Note that `$IsWindows` is not set on Windows PowerShell (that is, versions 5.1 and earlier). In general, when testing `$IsWindows`, it's helpful to also look at the version table in order to support older versions.
```suggestion
    if ($IsWindows -or $PSVersionTable.PSEdition -eq "Desktop") {
```

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
+        $old_PATH = $env:PATH
+        if (!(Get-Command clang -ErrorAction SilentlyContinue)) {
+            if ((choco find --idonly -l llvm) -notcontains "llvm") {
+                Write-Host "##[info]Installing LLVM with Chocolatey."
+                choco install llvm --version=11.1.0
```
cgranade, line 149: Similarly here, this makes a global and possibly disruptive change to the caller's environment.

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
+            }
+
+            if ((choco find --idonly -l llvm) -contains "llvm") {
```
cgranade, line 152: This runs `choco find --idonly -l llvm` twice... as a possible optimization, you may want to change this to an `else` block.

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
+    }
 } else {
+    Write-Host "##[info]The Microsoft.Quantum.Qir.CommandLineTool is already installed."
+}
+
+if ($allOk) {
+    if ($IsWindows) {
+        $old_PATH = $env:PATH
+        if (!(Get-Command clang -ErrorAction SilentlyContinue)) {
+            if ((choco find --idonly -l llvm) -notcontains "llvm") {
```
cgranade, line 147: Is chocolatey required? If so, it may be good to check that explicitly earlier.

Build/validate-qir.ps1
```diff
@@ -89,35 +104,71 @@ $qirProjects = @(
...
+            } else {
+                throw "Unable to find an installation of LLVM."
+            }
+        }
+    }
+
+    Write-Host "##[info]Running Validation of QIR against Samples."
     $qirProjects `
         | ForEach-Object { Build-One $_.Path $_.Args }
+    if ($IsWindows) {
```
cgranade, line 164: Similarly here, may want to guard on checking `$PSVersionTable.PSEdition` and to have the `$Env:PATH` assignment in a `finally`.

cgranade (commented):
I took a look through, and I think the biggest concern that jumped out at me is that the validation script now makes global changes to the caller's environment, in particular installing llvm and qir-cli if either or both is missing. That could be surprising for developers or contributors expecting the script to only perform validations, especially since installing LLVM is somewhat of a heavyweight change to make.

Build/validate-qir.ps1
ScottCarda-MS: @swernli had suggested that $IsWindows would be sufficient here. Since I'm not a powershell expert, I'll have to follow the lead of others on this.

Build/validate-qir.ps1
ScottCarda-MS: I'm not an expert on powershell, but the issue you raise doesn't seem to be the case here. The script worked correctly on Linux, Mac, and Windows.

Build/validate-qir.ps1
ScottCarda-MS: On Windows, the os is smart enough to look in the executable's location for library files, as well as PATH locations. Linux and Mac needed a little more help here, as they don't have that behavior.

Build/validate-qir.ps1
ScottCarda-MS: I noticed this as well, part way through doing this. I'm not sure what the reason is for the change in case for parameter names. @cesarzc might know more.

Build/validate-qir.ps1
ScottCarda-MS: When working on this with @anpaz, he suggested it would be best if the installation process was merged into the same script as the validation script. Following his suggestion, this script installs the resources necessary to perform it task of validation.

Build/validate-qir.ps1
ScottCarda-MS: Not that I've found. If it does become important in the future, the script can be updated at that time.

Build/validate-qir.ps1
cgranade: It absolutely is on PowerShell Core (6.0) and on PowerShell (7.0) or later, but it is not sufficient for Windows PowerShell (5.1 and earlier). In other PRs, you have indicated that you don't want to assume that 6.0 or later is installed. If you do want to take a hard dependency on 6.0 or later, adding the line `#Requires -Edition Core` at the top of the file (below the `#!` line) will enforce that.

https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_requires

Build/validate-qir.ps1
ScottCarda-MS: This is intentional. The first time is to check if it should do the install. The second is to check that the install was successful, and if so, proceed with the validation.

Build/validate-qir.ps1
cgranade: I think that can make sense when installation procedures are local in nature (e.g. restoring a csproj or manifest), but I'm more wary of modifying global environments from an integration test script.

Build/validate-qir.ps1
ScottCarda-MS: Chocolatey is only required if we are on Windows and `clang` is not found.

Build/validate-qir.ps1
cgranade: A comment indicating that may be helpful, then. As written, it's not clear that the second call is to check for errors.

Build/validate-qir.ps1
ScottCarda-MS: Let's consolidate this conversation into the related comment thread. Both if-statements will have the same condition regardless of what that condition is.

Build/validate-qir.ps1
ScottCarda-MS: done

Build/validate-qir.ps1
ScottCarda-MS: done

Build/validate-qir.ps1
ScottCarda-MS: Yes this issue here: https://github.com/microsoft/qsharp-runtime/issues/757. I've added this link to the comments, per your suggestion.

Build/validate-qir.ps1
cgranade: That behavior is configurable on Linux and macOS, I believe, by setting `rpath` to `$ORIGIN` when compiling the executable.

Build/validate-qir.ps1
ScottCarda-MS: added

Build/validate-qir.ps1
bettinaheim: No reason, just a matter of us properly aligning all APIs.

Build/validate-qir.ps1
bettinaheim: Making the edit that Chris suggests here makes sense.

ScottCarda-MS: done

Build/validate-qir.ps1
bettinaheim: The cli tool at this time is intentionally kept internal and the way to get it is to download it from the build. I also don't expect that we have many changes to the command line itself. I hence think it is fine as is for now.

Build/validate-qir.ps1
bettinaheim: This seems like a larger design decision for how our build and build scripts work. Let's make that a separate conversation to have when all the people that are involved with builds are back. The way it is set up now is aligned with Andres' suggestions and matches what is also done on other repos. There is certainly value in pursuing a different setup, but I think then we need to schedule time to make a pass over all our repos and refactor some build scripts.

Build/validate-qir.ps1
bettinaheim: See my comment below on the same topic.

bettinaheim (approved):
Let's pull in the initial setup and have a broader conversation about the design of our build setup across repos separately.

ScottCarda-MS:
I've merged in the PR as it is and left the open comments up for us to follow up on later. We should have a broader conversation about the things raised here and how we want to implement them in our build scripts.

PR #524, opened 2021-07-20
Integration test for AutoSubstitution rewrite step

msoeken:
This adds an integration test for the auto substitution sample to check the rewrite step in builds.

### DIFF
--- a/Build/test.ps1
+++ b/Build/test.ps1
@@ -38,10 +38,29 @@ function Validate-Integrals {
     }
 }
 
+function Test-AutoSubstitution {
+    Write-Host "##[info]Testing AutoSubstitution sample..."
+
+    $output = dotnet run --project (Join-Path $PSScriptRoot "../samples/runtime/autosubstitution")
+
+    if ($output -notmatch "^Quantum version\s*$") {
+        Write-Host "##vso[task.logissue type=error;]Auto substitution with QuantumSimulator failed"
+        $script:all_ok = $False
+    } else {
+        $output = dotnet run --project (Join-Path $PSScriptRoot "../samples/runtime/autosubstitution") -s ToffoliSimulator
+
+        if ($output -notmatch "^Classical version\s*$") {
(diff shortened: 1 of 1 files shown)


Build/test.ps1
```diff
@@ -38,10 +38,29 @@ function Validate-Integrals {
     }
 }
 
+function Test-AutoSubstitution {
```
cgranade, line 41: Since it's not generally a scalable approach to have a different cmdlet for each sample being tested, it'd be helpful to explain in comments why autosubstitution is unusual in that regard. Hopefully, that will reduce the temptation to make one-off cmdlets for other samples in the future.

Build/test.ps1
```diff
@@ -38,10 +38,29 @@ function Validate-Integrals {
...
+    Write-Host "##[info]Testing AutoSubstitution sample..."
+
+    $output = dotnet run --project (Join-Path $PSScriptRoot "../samples/runtime/autosubstitution")
+
+    if ($output -notmatch "^Quantum version\s*$") {
+        Write-Host "##vso[task.logissue type=error;]Auto substitution with QuantumSimulator failed"
```
cgranade, line 47: Can we add a bit more detail to the error log as to what the failure is? E.g.: make the message say that the wrong version of the operation was executed?

Build/test.ps1
```diff
@@ -38,10 +38,29 @@ function Validate-Integrals {
...
+        $script:all_ok = $False
+    } else {
+        $output = dotnet run --project (Join-Path $PSScriptRoot "../samples/runtime/autosubstitution") -s ToffoliSimulator
+
+        if ($output -notmatch "^Classical version\s*$") {
+            Write-Host "##vso[task.logissue type=error;]Auto substitution with ToffoliSimulator failed"
```
cgranade, line 53: Same comment here.

cgranade (approved):
Looks good to me, just had a couple quick comments.

PR #525, opened 2021-07-20
Fix typos across multiple samples

tcNickolas:


cgranade (approved):
Thanks for catching these! Perhaps including spell check warnings would be a good thing to enable in CI...

tcNickolas:
Agreed, but we'll have to do a lot of work teaching the spellchecker our terminology, names, and variable names - my spellchecker showed several thousands of warnings, I had to filter them quite strictly

cgranade:
> Agreed, but we'll have to do a lot of work teaching the spellchecker our terminology, names, and variable names - my spellchecker showed several thousands of warnings, I had to filter them quite strictly

Put together a quick attempt at #526, as I've had good luck with cspell before. I agree there's likely to be a lot of false positives, so set it to warnings only for now.

PR #526, opened 2021-07-20
Check spelling on PRs.

cgranade:
This PR adds the [`cspell-action` GitHub Action](https://github.com/marketplace/actions/cspell-action) as a check on changed files in PRs. Since it is likely that we'll get a lot of false positives initially, this PR sets spell checks to warning levels only for now.


cspell.json
```diff
@@ -0,0 +1,29 @@
...
+        // Terms that come up in build / CI configuration.
+        "logissue", // Used in ##[vso.logissue] commands.
+        "pwsh",
+        "markdownlint",
+
+        // Words specific to quantum computing.
+        "eigenstate",
+        "qubit",
+        "qubits",
+        "qubitization",
```
guenp, line 16:
```suggestion
        "qubit's",
        "qubitization",
        "Chuang",
```

cspell.json
```diff
@@ -0,0 +1,29 @@
...
+
+        // Chemistry terminology
+        "nitrogenase",
+        "Hartree",
+        "Fock"
+    ],
+
+    "flagWords": [
+        // The Trotter–Suzuki decomposition isn't a verb.
+        "Trotterization"
```
guenp, line 27: is this personal preference?

guenp (approved):
This is neat, thanks for adding!

cspell.json
cgranade: Not per se, I don't think, so much as that examples of denominal verbs formed from proper nouns that refer to people are generally used to indicate "act in the manner of" rather than "use the invention of" (i.e.: ran across this neat paper that surveyed different semantic categories of denominal verbs formed from proper nouns: https://journals.openedition.org/lexis/4681#tocfrom2n8), such that it's much more unambiguous to explicitly state "use the invention of" (in this case, "use the Trotter–Suzuki decomposition").

There's also that the Trotter decomposition is a special case of the lowest-order Trotter–Suzuki decomposition. In most cases, we don't restrict to the lowest-order decomposition.

PR #528, opened 2021-07-21
Disable Test-AutoSubstitution

anpaz:


Build/test.ps1
```diff
@@ -64,7 +64,9 @@ Validate-Integrals
 
 Test-One '../samples/tests/sample-tests'
 Test-One '../samples/diagnostics/unit-testing'
-Test-AutoSubstitution
+
+# Disabled for now, as it breaks in e2e builds
```
cgranade, line 68: Is there an environment variable we can condition on so that this still runs in per-repo and local tests?

cgranade (approved):

PR #532, opened 2021-08-02
Update Coulomb repulsion, SCF energy

### LINKED ISSUE #506: H2 Broombridge file seems incorrect
**Describe the bug**
@dgwvfxn pointed out that the Broombridge file for Hydrogen included in one of the Chemistry samples contains the incorrect SCF energy: [h2.yaml](https://github.com/microsoft/Quantum/blob/main/samples/chemistry/PythonIntegration/h2.yaml). He has generated a new file that should contain the correct values.

The values for the coulomb repulsion and SCF energy should be:
```
  coulomb_repulsion:
    value: 0.32584805972887831
    units: hartree
  scf_energy:
    value: -0.87499072
    units: hartree
```

Another error is the distance between H atoms: it is given in Bohr, not Angstrom.

guenp:
Fixes #506 

CC @dgwvfxn


samples/chemistry/PythonIntegration/h2.yaml
```diff
@@ -7,14 +7,14 @@ format: {version: '0.1'}
 generator: {source: nwchem, version: '6.8'}
```
cgranade, line 7: Can we add `nwchem` to the allowed words in `cspell.json` to suppress this warning? It may also be good to add `"Mulliken"` to `cspell.json` too, since that's used fairly commonly used in the chemistry stuff, and since it's also getting flagged below.

samples/chemistry/PythonIntegration/h2.yaml
```diff
@@ -7,14 +7,14 @@ format: {version: '0.1'}
...
-  coulomb_repulsion: {units: hartree, value: 0.713776188}
+  coulomb_repulsion: {units: hartree, value: 0.32584805972887831}
   energy_offset: {units: hartree, value: 0.0}
   fci_energy: {lower: 0.0, units: hartree, upper: 0.0, value: 0.0}
   geometry:
     atoms:
     - coords: [0.0, 0.0, 0.0]
       name: H
-    - coords: [0.0, 0.0, 1.624]
+    - coords: [0.0, 0.0, 0.859]
```
cgranade, line 17: It looks like the bond length has been changed in this PR as well. That should be perfectly fine since there's no particular "right" bond length to use, but out of curiousity, do we know if the old Coulomb repulsion and SCF energy were consistent for the old bond length?

cgranade (approved):

samples/chemistry/PythonIntegration/h2.yaml
guenp: The bond length is the same, I just made it consistent with the units (angstrom).

PR #535, opened 2021-08-12
Convert algorithm samples to standalone executables to enable QIR testing

samarsha:
Replaced the C# and Python drivers with standalone Q# executables for these algorithm samples, so that the QIR validation script in the QDK repo can run them:

* CHSH Game
* Database Search

I didn't update these algorithm samples which are still incompatible with QIR:

* Integer Factorization. It uses a custom simulator in C#, which I don't think is supported by QIR.
* Sudoku Solver. In principle, I think it might be possible to port this C# code to Q#, but there's so much of it, and it uses things like hash sets that Q# doesn't have a good representation of.


samples/algorithms/chsh-game/Program.qs
```diff
@@ -0,0 +1,77 @@
...
+        QuantumWins : Int
+    );
+
+    /// # Summary
+    /// Compares the success rates of the classical and quantum CHSH game strategies across 10,000 trials.
+    @EntryPoint()
+    operation CompareStrategies() : Unit {
+        let trials = 10000;
+        let scores = DrawMany(PlayGame, trials, ());
+        let total = Fold(PlusScore, Score(0, 0), scores);
```
cgranade, line 29: I really love this way of writing out how to run lots of trials, but it may be unfamiliar to new Q# users such that adding some comments about what `DrawMany` and `Fold` are doing here could be helpful.

samples/algorithms/chsh-game/README.md
```diff
@@ -67,8 +57,7 @@ Press Start in Visual Studio to run the sample.
 ## Manifest
 
 - [Game.qs](https://github.com/microsoft/Quantum/blob/main/samples/algorithms/chsh-game/Game.qs): Q# code implementing the game.
-- [host.py](https://github.com/microsoft/Quantum/blob/main/samples/algorithms/chsh-game/host.py): Python host program to call into the Q# sample.
-- [Host.cs](https://github.com/microsoft/Quantum/blob/main/samples/algorithms/chsh-game/Host.cs): C# code to call the operations defined in Q#.
+- [Program.qs](https://github.com/microsoft/Quantum/blob/main/samples/algorithms/chsh-game/Program.qs): Q# program entry point for the example.
```
cgranade, line 60: It looks like this is the link that's flagging the error in link checking, but it should be valid once your PR is completed such that I think we should be OK to ignore that optional check.

samples/algorithms/database-search/Program.qs
```diff
@@ -0,0 +1,181 @@
...
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Convert;
+
+    @EntryPoint()
+    operation RunDatabaseSearches() : Unit {
+        RunRandomSearch();
+        RunQuantumSearch();
+        RunMultipleQuantumSearch();
+    }
```
cgranade, line 14: Would this be a good place to demonstrate the multiple entry-point feature?

samples/algorithms/database-search/Program.qs
```diff
@@ -0,0 +1,181 @@
...
+            "\n\nQuantum search for marked element in database.\n" +
+            $"  Database size: {databaseSize}.\n" +
+            $"  Marked elements: {markedElements}" +
+            $"  Classical success probability: {classicalSuccessProbability}\n" +
+            $"  Queries per search: {queries} \n" +
+            $"  Quantum success probability: {quantumSuccessProbability}\n"
+        );
+
+        for attempt in 1 .. repeats {
+            // Extract the marked qubit state
```
cgranade, line 146:
```suggestion
            // Extract the marked qubit state.
```

samples/algorithms/database-search/Program.qs
```diff
@@ -0,0 +1,181 @@
...
+            let (markedQubit, databaseRegister) = ApplyGroverSearch(markedElements, nIterations, nDatabaseQubits);
+            set successCount += markedQubit == One ? 1 | 0;
+
+            // Print the results of the search every attempt
```
cgranade, line 150:
```suggestion
            // Print the results of the search every attempt.
```

cgranade (approved):
Looks good to me, just had a few minor comments.

samples/algorithms/chsh-game/README.md
samarsha: I changed the links to be relative, which fixed the error.

samples/algorithms/chsh-game/Game.qs
```diff
@@ -143,4 +141,22 @@ namespace Microsoft.Quantum.Samples.CHSHGame {
             return MeasureBobQubit(bobBit, bobQubit) == MeasureAliceQubit(aliceBit, aliceQubit);
         }
     }
+
+    /// # Summary
+    /// Plays the optimal classical strategy for the CHSH game: both Alice and Bob both always
+    /// output 0. This should result in success 75% of the time.
```
kuzminrobin, line 147: FYI: "**both** Alice and Bob **both** always".
(one **both** seems to me redundant)

samples/algorithms/chsh-game/Game.qs
```diff
@@ -143,4 +141,22 @@ namespace Microsoft.Quantum.Samples.CHSHGame {
...
+    /// ## bobBit
+    /// The bit given to Bob (Y).
+    ///
+    /// # Output
+    /// Whether Alice and Bob's output bits match.
+    function PlayClassicalStrategy(aliceBit : Bool, bobBit : Bool) : Bool {
+        let aliceOutput = false;
+        let bobOutput = false;
+        return aliceOutput == bobOutput;
+    }
```
kuzminrobin, line 161: (Just Curiosity)
> function PlayClassicalStrategy(**aliceBit** : Bool, **bobBit** : Bool) : Bool

Doesn't compiler complain that the parameters are not used in the function? If not, then is there a chance that in the the future the compiler or static analyzer will?

In C/C++ if the parameter is not used then it is OK to skip the param's name in the param list (to calm down the compiler/analyzer). Is the same possible in .qs?
> function PlayClassicalStrategy( **: Bool, : Bool**) : Bool

kuzminrobin (approved):

samples/algorithms/chsh-game/Game.qs
samarsha: > Doesn't compiler complain that the parameters are not used in the function? If not, then is there a chance that in the the future the compiler or static analyzer will?

The Q# compiler doesn't do this currently. I think it would make sense to add that feature though.

> In C/C++ if the parameter is not used then it is OK to skip the param's name in the param list (to calm down the compiler/analyzer). Is the same possible in .qs?

The `: Bool` syntax isn't valid, but what is typically done in ML-style languages like Q# is to prefix or replace the name with an underscore if it's unused. For example, if `aliceBit : Bool` was unused and generated a warning, renaming it to `_aliceBit : Bool` or `_ : Bool` would silence the warning. (Although `_` is not currently allowed in Q# as a parameter name, I think it would make sense to lift that restriction.)

samples/algorithms/chsh-game/Program.qs
samarsha: I added comments for these lines. Do they help?

cgranade:
Yep, thank you!

________________________________

samples/algorithms/database-search/Program.qs
samarsha: I updated it to use multiple entry points now. Could you check if the changes to the readme are clear enough? (I left the IQ# version alone since it seems to define its own entry point.)

samples/algorithms/database-search/README.md
```diff
@@ -33,20 +38,19 @@ This sample walks through Grover's search algorithm. Oracles implementing the da
...
+### Visual Studio Code or the Command Line
 
-At a terminal, run the following command:
+At a terminal, run the following commands for each of the three examples:
 
 ```powershell
-python host.py
-```
+# Example 1:
+dotnet run simulate Microsoft.Quantum.Samples.DatabaseSearch.RunRandomSearch
```
cgranade, line 48: [nit] for usage on mybinder.org, splitting into multiple Markdown blocks allows for this Markdown file to be used as a command-line notebook to run each sample independently.
````suggestion
```

```
````

cgranade (approved):

PR #547, opened 2021-09-10
New Q# + Python and process tomography samples.

cgranade:
This PR moves the process tomography sample out from the interoperability section, ports it to use the open systems simulator, and adds a new sample focused on Q# + Python workflow features such as `%%qsharp` and diagnostics capture.


msoeken (approved):
Thanks for the sample, looks good to me.

PR #549, opened 2021-09-11
Updated the QDK version to 0.18.2109.162713

kuzminrobin:


kuzminrobin:
The tag and the branch created based on that tag are based on a too early commit in history, that is why this PR conflicts with main. Closing in favor of #550.

PR #558, opened 2021-09-30
Draft of new GST sample.

cgranade:
This PR adds a new gateset tomography sample, complementing the existing process tomography and randomized benchmarking samples. In particular, this sample uses the pyGSTi package to infer noise models on a two-qubit {X, Y, CNOT} gateset.

Still in draft pending the following work items:
- [ ] Add README
- [ ] Link from binder-index.
- [ ] Add short introduction to GST.
- [ ] Lock Q# version in environment.yml.


samples/characterization/gateset-tomography/gateset-tomography.ipynb
```diff
@@ -0,0 +1,1850 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0fa6fef9",
+   "metadata": {},
+   "source": [
+    "# Gateset Tomography with Python and Q# #"
```
msoeken, line 8:
```suggestion
    "# Gateset Tomography with Python and Q#"
```
I think the recommendation for markdown headers is to only have the hash symbol in the beginning of the line.

samples/characterization/gateset-tomography/gateset-tomography.ipynb
```diff
@@ -0,0 +1,1850 @@
...
+    "    max_lengths\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "08b3a978",
+   "metadata": {},
+   "source": [
+    "Each of these experiments consists of running a nonadaptive quantum program (colloquially, a \"circuit\") a number of times and collecting the results."
```
msoeken, line 142: Maybe one can write _quantum circuit_ and then remove the quotes.

msoeken (commented):
I am leaving some comments, since the PR is still in draft mode.

samples/characterization/gateset-tomography/gateset-tomography.ipynb
cgranade: Agreed, this was a bit of a hack, since the heading ends in a "#." The problem is not all clients support escaping the hash in Q#, so this is the safest way to typeset "Q#" at the end of a header...

msoeken: True, that makes sense. Then we better leave it there :)

PR #562, opened 2021-10-07
Some changes to README and CMakeLists in oracle generator

msoeken:


msoeken:
> Not a blocking issue, but could be good to add `mkdir` and similar to `cspell.json` to reduce false positives.

Hi @cgranade, I am afraid I don't know what you are referring to. Could you please elaborate a bit more?

PR #580, opened 2021-11-29
Update dotnet version to fix docker build

anjbur:


Build/images/samples/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get -y update && \
                libpng-dev \
                # iqsharp-base currently ships with .NET Core SDK 3.1, but
                # we need .NET 5.0 to run dotnet-interactive.
```
cgranade, line 25: Nit, but this comment may need to be updated as well?

cgranade (approved):

Build/images/samples/Dockerfile
anjbur: Oh good catch, thank you! `dotnet-interactive` is the package requiring the update to .NET 6.0, so this should definitely be updated.

PR #581, opened 2021-11-29
Update Samples version to use Nov 21 release 0.21.2111.177148

guenp:


samples/algorithms/variational-algorithms/environment.yml
```diff
@@ -8,7 +8,7 @@ channels:
 dependencies:
   - notebook
   - matplotlib
-  - qsharp==0.20.2110171573
+  - qsharp==0.21.2111177148
```
cgranade, line 11: Did you get a chance to double-check if this environment creates successfully? I may need to add another CI job that does a `conda env create --dry-run` on each envionment.yml file eventually...

cgranade (approved):

samples/algorithms/variational-algorithms/environment.yml
guenp: That won't work, it's missing a dot! Let me fix that.

cgranade: Thanks! We ran into a previous update incorrectly removing some dots and not getting caught by CI so wanted to quickly double check. Thanks for fixing that, really appreciate it!

PR #582, opened 2021-12-08
Resources estimation with quantum-viz.js

msoeken:
This extends the integer factorization sample to produce JSON output for quantum-viz.js, incl. the hierarchy of the call stack and annotated resource estimates to each operation in the call stack.

### DIFF
--- a/samples/algorithms/integer-factorization/Program.cs
+++ b/samples/algorithms/integer-factorization/Program.cs
@@ -49,8 +49,11 @@ namespace Microsoft.Quantum.Samples.IntegerFactorization
         [Option('g', "generator", Required = true, HelpText = "A coprime to `number` of which the period is estimated")]
         public long Generator { get; set; }
 
-        [Option('r', "resource", Required = false, Default = 0, HelpText = "The resource - CNOT: 0; Measure: 1; QubitClifford: 2; R: 3; T: 4")]
+        [Option('r', "resource", Required = false, Default = 0, HelpText = "The resource - CNOT: 0; QubitClifford: 1; Measure: 2; R: 3; T: 4")]
         public PrimitiveOperationsGroups Resource { get; set; }
+
+        [Option("quantum-viz", HelpText = "Use quantum-viz.js for visualization")]
+        public bool QuantumViz { get; set; }
     }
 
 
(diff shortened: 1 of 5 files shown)


samples/algorithms/integer-factorization/Program.cs
```diff
@@ -49,8 +49,11 @@ class VisualizeOptions : CommonOptions
         [Option('g', "generator", Required = true, HelpText = "A coprime to `number` of which the period is estimated")]
         public long Generator { get; set; }
 
-        [Option('r', "resource", Required = false, Default = 0, HelpText = "The resource - CNOT: 0; Measure: 1; QubitClifford: 2; R: 3; T: 4")]
+        [Option('r', "resource", Required = false, Default = 0, HelpText = "The resource - CNOT: 0; QubitClifford: 1; Measure: 2; R: 3; T: 4")]
```
tcNickolas, line 52: Is there any place in the code that needs a matching change?

samples/algorithms/integer-factorization/Program.cs
```diff
@@ -142,14 +145,25 @@ static int Estimate(EstimateOptions options)
...
-            var config = FlameGraphResourcesEstimator.RecommendedConfig();
+            if (options.QuantumViz) {
+                var config = QuantumVizEstimator.RecommendedConfig();
 
-            var estimator = new FlameGraphResourcesEstimator(config, options.Resource);
+                var estimator = new QuantumVizEstimator(config);
 
-            var bitsize = (long)System.Math.Ceiling(System.Math.Log2(options.NumberToFactor + 1));
-            EstimateFrequency.Run(estimator, options.Generator, options.NumberToFactor, options.UseRobustPhaseEstimation, bitsize).Wait();
+                var bitsize = (long)System.Math.Ceiling(System.Math.Log2(options.NumberToFactor + 1));
```
tcNickolas, line 153: This line can be moved outside of the if statement, it doesn't seem to depend on the path

tcNickolas (commented):
I left a couple of small comments, but I'll need to read the updated blog to make sure I understand what the code is doing

samples/algorithms/integer-factorization/Program.cs
msoeken: Thanks for flagging this. There is no code affected by this, it was just wrong documentation in the CLI info text.

samples/algorithms/integer-factorization/QuantumVizCounter.cs
```diff
@@ -0,0 +1,200 @@
...
+        [JsonProperty(PropertyName = "displayArgs")]
+        public string DisplayArgs { get => $"CNOT: {CNOTCount} Clifford: {CliffordCount} R: {RCount} T: {TCount}"; }
+        [JsonProperty(PropertyName = "children")]
+        public List<Operation> Children { get; set; } = new List<Operation>();
+        [JsonProperty(PropertyName = "targets")]
+        public HashSet<QubitReference> Targets { get; set; } = new HashSet<QubitReference>();
+        [JsonProperty(PropertyName = "isAdjoint")]
+        public bool IsAdjoint { get; set; } = false;
+
+        [JsonIgnore]
```
tcNickolas, line 143: Minor: I was initially confused as for why we're ignoring these fields in JSON generation if these are the fields we're looking for - I'd leave a comment here, something like "// We're not outputting the resource count fields in the JSON, since we include them in DisplayArgs field"

tcNickolas (approved):
Looks good to me!

PR #585, opened 2021-12-16
QDK Dec 2021 release

vxfield:


samples/algorithms/variational-algorithms/environment.yml
```diff
@@ -8,7 +8,7 @@ channels:
 dependencies:
   - notebook
   - matplotlib
-  - qsharp==0.21.2111.177148
+  - qsharp==0.21.2112180703
```
cgranade, line 11: It looks like the extra dot got dropped here? Please double-check all four environment.yml files to make sure that they work with the new package versions. Thank you!

cgranade (approved):
Approved with one comment that should be addressed before merging. Thanks!

samples/algorithms/variational-algorithms/environment.yml
vxfield: Thanks, Chris.
I have also fixed the Bash script to correctly format the Nuget, Python and Docker versions regardless of how the input version is given.

vxfield: I'll also update the QDK repo (which may be overriding this script) with the correct formatting of versions.

PR #586, opened 2021-12-20
Update QDK to .NET6.0 (Quantum)

ricardo-espinoza:
We're migrating the QDK to the most recent Long Time Support version of the .NET framework. For details about this change, refer to the original issue [Q#C:1224](https://github.com/microsoft/qsharp-compiler/issues/1224).

As part of this change, we're:

- Re-targeting all .NetCoreApp3.1 binaries to .NET6.0
- Updating Docker images, samples and templates.
- Libraries using .NetStandard2.1 are not affected by this change.
- The minimum supported .NET version in the QDK will also be updated from 3.1 to 6.0

This is a multi-repo change, and all changes will need to be merged concurrently.


samples/algorithms/variational-algorithms/environment.yml
```diff
@@ -8,7 +8,7 @@ channels:
 dependencies:
   - notebook
   - matplotlib
-  - qsharp==0.21.2112.180703
+  - qsharp==0.21.2112182074b1
```
msoeken, line 11: Is this version number correct? I am just double checking, because it looks different from the others.

msoeken (approved):

samples/algorithms/variational-algorithms/environment.yml
ricardo-espinoza: Yes, but only while we keep this as a beta release.

PR #621, opened 2022-03-22
Update repo CI builds to use .NET6

ricardo-espinoza:


tcNickolas (approved):
Looks good.
Markdown failures seem to be a false positive, I checked that both reported links are fine.

PR #625, opened 2022-03-24
move gh-sync to workflows

IsraelMiles:
The gh-sync.yml file needs to be moved to workflows, my apologies for the wrong location.


cgranade (approved):
My apologies, I should have caught this as well!

PR #638, opened 2022-03-31
Fix more sample notebook issues

achocron:
Found a few more issues in sample notebooks.

### DIFF
--- a/samples/algorithms/noisy-dj/Deutsch–Jozsa with Noise.ipynb
+++ b/samples/algorithms/noisy-dj/Deutsch–Jozsa with Noise.ipynb
@@ -513,14 +513,6 @@
    "source": [
     "qsharp.component_versions()"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "industrial-policy",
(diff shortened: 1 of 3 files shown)


cgranade (commented):
As a high-level remark, please make sure to apply changes to all places that samples appear, not just for the portal gallery.

achocron:
@cgranade fair enough - but note that this is being merged into the `feature/samples-gallery` branch rather than `main`, so I didn't spend time going through all of the other samples here. I can open another PR to go into main for general touch ups like these, though.

cgranade:
> @cgranade fair enough - but note that this is being merged into the `feature/samples-gallery` branch rather than `main`, so I didn't spend time going through all of the other samples here.

To clarify, I'm not suggesting going through any other samples, but rather that we need to make sure to avoid unnecessary divergence between `main` and `feature/sample-gallery`. From that perspective, I would greatly appreciate a matching PR against `main` following this one. Thanks!

xinyi-joffre (dismissed):

xinyi-joffre (approved):

PR #661, opened 2022-04-25
Update npm packages

msoeken:
This applies some recommended updates to NPM packages.


cgranade:
Have you tested these changes manually to ensure that they do not break the respective samples? Changes to chart.js can cause GUI samples to stop working such that manual validation is generally required. Thanks!

PR #680, opened 2022-06-14
WIP to improve test stability

achocron:


.github/workflows/hosted-notebook-samples-check.yml
```diff
@@ -4,27 +4,62 @@ on:
...
+    outputs:
+      download_urls: ${{ steps.prepare.outputs.urls }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Prepare variables
+        id: prepare
+        run: |
+          FILES=$(git diff origin/${{ github.event.pull_request.base.ref }} --name-only)
```
cgranade, line 19: I think I'm missing a little context, sorry; what variables is this script preparing, and how do those get used by later steps? A few comments explaining the intent and the flow here would be helpful, perhaps?

achocron: Ah yes, I suppose that would be helpful :) basically this is looking at the files that have been changed by the PR relative to the base branch. Then, for each *.ipynb, the raw file URL within the current commit is stored in a variable. This variable is sent to pipeline in azure devops, where each file url is tested as if it were an existing notebook sample, even if it isn't. Then, if the cells all run without errors for each notebook, the pipeline passes.

cgranade: Ah, ok, so this allows restricting tests in hosted notebooks to only those notebooks changed by a given PR?

achocron: Exactly!

xinyi-joffre (approved):

PR #687, opened 2022-06-27
Adapt samples to API changes in open systems simulator.

cgranade:


samples/characterization/process-tomography/tomography-sample.ipynb
```diff
@@ -123,7 +121,18 @@
    "cell_type": "code",
    "execution_count": 5,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
```
anjbur, line 126: Is it intended that there be an error in the output here?

samples/characterization/randomized-benchmarking/randomized-benchmarking.ipynb
```diff
@@ -126,7 +123,18 @@
    "cell_type": "code",
    "execution_count": 5,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
```
anjbur, line 128: Same question here, about if the error is intended

anjbur (approved):
I had a question about error output in two cells, but otherwise looks good to me!

samples/characterization/randomized-benchmarking/randomized-benchmarking.ipynb
cgranade: Same here as well; these could be resolved by fixes to QInfer.

cgranade (commented):
Thanks for your review, @anjbur!

PR #688, opened 2022-06-28
Added storage sample

anraman:
Added storage/data management sample showing users how to manage data in the notebooks experience through the portal

### DIFF
--- a/samples/azure-quantum/utilities/storage-data-management.ipynb
+++ b/samples/azure-quantum/utilities/storage-data-management.ipynb
@@ -1,0 +1,758 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "nteract": {
+          "transient": {
+            "deleting": false
+          }
+        }
+      },
+      "source": [
+        "# Data Management Sample\n",
+        "This sample shows you how to connect and use Azure Quantum with external datasources such as Azure Blob Storage. "
+      ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


samples/azure-quantum/utilities/storage-data-management.ipynb
```diff
@@ -0,0 +1,758 @@
...
+        "nteract": {
+          "transient": {
+            "deleting": false
+          }
+        }
+      },
+      "outputs": [],
+      "source": [
+        "# Log in to the Azure CLI (only need to do once per session)\n",
+        "!az login"
```
xinyi-joffre, line 91: @anpaz, is there any way to use the existing azure-quantum libs to connect to the linked storage account without doing the !az login method in this sample?

I found this way to get a blob, but it seems a bit hacky:

from azure.quantum import Workspace
workspace = Workspace (
    subscription_id = "677fc922-91d0-4bf6-9b06-4274d319a0fa",
    resource_group = "xiou",
    name = "xiou-notebooks-demo",
    location = "eastus2euap"
)

from azure.quantum._client.models import BlobDetails
sas_uri_response = workspace._client.storage.sas_uri(blob_details=BlobDetails(container_name="my-container"))

from azure.storage.blob import ContainerClient
container_client = ContainerClient.from_container_url(container_url=sas_uri_response.sas_uri)
container_client.exists()

anraman:
Will add as a docs page instead

PR #695, opened 2022-07-13
Optimize Grover's Search Sudoku Sample

adrianleh:
This pull request optimizes the Grover's search Sudoku sample by encoding constraints into search space minimizing required quantum computation and increasing 0 component of state space.

This change allows enables us to also switch over to the sparse simulator ultimately resulting in greatly increased performance. We now have the ability to solve the puzzle [1] on a regular laptop correctly within about 45min requiring 150 iterations, whereas the previous version estimates to require 804 iterations and does not complete in reasonable time.

The difference in [2] is even more significant, the proposed change solves the puzzle instantly with 0 iterations, whereas the current version estimates to require 205887 iterations.

[1]: 
```csharp
{
{ 0,7,3, 8,0,0, 5,1,2 },
{ 9,0,2, 7,0,5, 4,8,6 },
{ 8,4,5, 0,0,0, 0,0,0 },
{ 7,0,8, 2,0,1, 3,5,0 },
{ 5,2,6, 4,0,3, 8,9,1 },
{ 1,3,4, 5,0,0, 0,0,0 },
{ 4,6,9, 1,2,8, 7,3,5 },
{ 2,8,7, 3,5,6, 1,4,9 },
{ 3,5,1, 9,4,7, 6,0,8}
};
```

[2]:

```csharp
{
{ 0,7,3, 8,9,4, 5,1,2 },
{ 9,0,2, 7,3,5, 4,8,6 },
{ 8,4,5, 6,1,2, 9,7,3 },
{ 7,0,8, 2,0,1, 3,5,0 },
{ 5,2,6, 4,0,3, 8,9,1 },
{ 1,3,4, 5,8,9, 0,0,0 },
{ 4,6,9, 1,2,8, 7,3,5 },
{ 2,8,7, 3,5,6, 1,4,9 },
{ 3,5,1, 9,4,7, 6,2,8}
};
```

### DIFF
--- a/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
+++ b/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
@@ -124,7 +124,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
         within {
             for ((start, end), conflictQubit) in Zipped(edges, edgeConflictQubits) {
                 // Check that endpoints of the edge have different colors:
-                // apply ColorEqualityOracle_Nbit oracle;
+                // apply ApplyColorEqualityOracle oracle;
                 // if the colors are the same the result will be 1, indicating a conflict
                 ApplyColorEqualityOracle(
                     colorsRegister[start * bitsPerColor .. (start + 1) * bitsPerColor - 1],
--- a/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
+++ b/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
@@ -150,7 +150,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ## bitsPerColor
     /// The number of bits per color.
     /// ## nIterations
-    /// An estimate of the maximum iterations needed.
+    /// The number of iterations needed.
     /// ## oracle
     /// The Oracle used to find solution.
     /// ## statePrep
--- a/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
+++ b/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
@@ -30,8 +30,8 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ## register
     /// The register of qubits to be measured.
     operation MeasureColoring (bitsPerColor : Int, register : Qubit[]) : Int[] {
-        let numVertices = Length(register) / bitsPerColor;
-        let colorPartitions = Partitioned(ConstantArray(numVertices - 1, bitsPerColor), register);
+        let nVertices = Length(register) / bitsPerColor;
+        let colorPartitions = Partitioned(ConstantArray(nVertices - 1, bitsPerColor), register);
         return ForEach(MeasureColor, colorPartitions);
     }
 
@@ -64,7 +64,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// non qubit vertices.
     ///
     /// # Input
-    /// ## numVertices
+    /// ## nVertices
     /// The number of vertices in the graph.
     /// ## bitsPerColor
     /// The bits per color e.g. 2 bits per color allows for 4 colors.
@@ -114,7 +114,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ```
     /// i.e. every vertex is connected to each other.
     operation ApplyVertexColoringOracle (
-        numVertices : Int, 
+        nVertices : Int, 
         bitsPerColor : Int, 
         edges : (Int, Int)[],
         colorsRegister : Qubit[],
@@ -145,7 +145,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// Using Grover's search to find vertex coloring.
     ///
     /// # Input
-    /// ## numVertices
+    /// ## nVertices
     /// The number of Vertices in the graph.
     /// ## bitsPerColor
     /// The number of bits per color.
@@ -154,23 +154,21 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ## oracle
     /// The Oracle used to find solution.
     /// ## statePrep
-    /// Routine that prepares an equal superposition of all basis states in the search space.
+    /// An operation that prepares an equal superposition of all basis states in the search space.
     ///
     /// # Output
-    /// Int Array giving the color of each vertex.
-    ///
+    /// An array giving the color of each vertex.
     operation FindColorsWithGrover(
-        numVertices : Int, bitsPerColor : Int, nIterations : Int,
+        nVertices : Int, bitsPerColor : Int, nIterations : Int,
         oracle : ((Qubit[], Qubit) => Unit is Adj),
         statePrep : (Qubit[] => Unit is Adj)) : Int[] {
 
         // Note that coloring register has the number of qubits that is
         // twice the number of vertices (bitsPerColor qubits per vertex).
-        use register = Qubit[bitsPerColor * numVertices];
+        use register = Qubit[bitsPerColor * nVertices];
 
         Message($"Trying search with {nIterations} iterations...");
         ApplyGroversAlgorithmLoop(register, oracle, statePrep, nIterations);
-        let res = MultiM(register);
         let coloring = MeasureColoring(bitsPerColor, register);
         ResetAll(register);
 
@@ -216,7 +214,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ## iterations
     /// The number of iterations to try.
     ///
-    /// # Output
+    /// # Remarks
     /// Unitary implementing Grover's search algorithm.
     operation ApplyGroversAlgorithmLoop(
         register : Qubit[],
--- a/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
+++ b/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
@@ -72,7 +72,7 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// the same color.
     ///
     /// # Output
-    /// An marking oracle that with the purpose of marking states as allowed where the colors of qubits related by an edge constraints are not equal.
+    /// An marking oracle that marks as allowed those states in which the colors of qubits related by an edge constraint are not equal.
     ///
     ///
     /// # Example
--- a/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
+++ b/samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
@@ -113,202 +113,32 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     ///  1---2
     /// ```
     /// i.e. every vertex is connected to each other.
-    /// Additionally, we require that:
-    ///
-    ///    - vertices 0 and 1 do not get colors 2 and 3.
-    ///    - vertices 2 and 3 do not get colors 3 and 0.
-    ///    - vertices 0 and 2 do not get colors 1 and 3.
-    ///    - vertices 1 and 3 do not get colors 2 and 0.
-    /// This results in edges (vertices that can not be same color):
-    /// `edges = [(1, 0),(2, 0),(3, 0),(3, 1),(3, 2)]`
-    /// This is saying that vertex 1 can not have same color as vertex 0 etc.
-    /// and startingColorConstraints = [(0, 1),(0, 3),(0, 2),(1, 2),(1, 0),
-    ///                    (1, 3),(2, 1),(2, 3),(2, 0),(3, 2),(3, 0),(3, 1)]
-    /// This is saying that vertex 0 is not allowed to have values 1,3,2
-    /// and vertex 1 is not allowed to have values 2,0,3
-    /// and vertex 2 is not allowed to have values 1,3,0
-    /// and vertex 3 is not allowed to have values 2,0,1
-    /// A valid graph coloring solution is: [0,1,2,3]
-    /// i.e. vertex 0 has color 0, vertex 1 has color 1 etc.
     operation ApplyVertexColoringOracle (
-        numVertices : Int, bitsPerColor : Int, edges : (Int, Int)[],
-        startingColorConstraints : (Int, Int)[],
+        numVertices : Int, 
+        bitsPerColor : Int, 
+        edges : (Int, Int)[],
         colorsRegister : Qubit[],
         target : Qubit
     )
     : Unit is Adj + Ctl {
         let nEdges = Length(edges);
-        let nStartingColorConstraints = Length(startingColorConstraints);
-        // we are looking for a solution that:
-        // (a) has no edge with same color at both ends and
-        // (b) has no Vertex with a color that violates the starting color constraints.
+        // we are looking for a solution that has no Vertex with a color that violates the starting color constraints.
         use edgeConflictQubits = Qubit[nEdges];
-        use startingColorConflictQubits = Qubit[nStartingColorConstraints];
         within {
-            ConstrainByEdgeAndStartingColors(
-                colorsRegister, edges, startingColorConstraints,
-                edgeConflictQubits, startingColorConflictQubits, bitsPerColor
-            );
-        } apply {
-            // If there are no conflicts (all qubits are in 0 state), the vertex coloring is valid.
-            ControlledOnInt(0, X)(edgeConflictQubits + startingColorConflictQubits, target);
-        }
-    }
-
-    operation ConstrainByEdgeAndStartingColors(
-        colorsRegister : Qubit[], edges : (Int, Int)[],
-        startingColorConstraints : (Int, Int)[],
-        edgeConflictQubits : Qubit[], startingColorConflictQubits : Qubit[], bitsPerColor: Int
-    )
-    : Unit is Adj + Ctl {
-        for ((start, end), conflictQubit) in Zipped(edges, edgeConflictQubits) {
-            // Check that endpoints of the edge have different colors:
-            // apply ColorEqualityOracle_Nbit oracle;
-            // if the colors are the same the result will be 1, indicating a conflict
-            ApplyColorEqualityOracle(
-                colorsRegister[start * bitsPerColor .. (start + 1) * bitsPerColor - 1],
-                colorsRegister[end * bitsPerColor .. (end + 1) * bitsPerColor - 1],
-                conflictQubit
-            );
-        }
-        for ((cell, value), conflictQubit) in Zipped(startingColorConstraints, startingColorConflictQubits) {
-            // Check that cell does not clash with starting colors.
-            ControlledOnInt(value, X)(
-                colorsRegister[cell * bitsPerColor .. (cell + 1) * bitsPerColor - 1],
-                conflictQubit
-            );
-        }
-
-    }
-
-    /// # Summary
-    /// Oracle for verifying vertex coloring, including color constraints
-    /// from non qubit vertices. This is the same as ApplyVertexColoringOracle,
-    /// but hardcoded to 4 bits per color and restriction that colors are
-    /// limited to 0 to 8.
-    ///
-    /// # Input
-    /// ## numVertices
-    /// The number of vertices in the graph.
-    /// ## edges
-    /// The array of (Vertex#,Vertex#) specifying the Vertices that can not
-    /// be the same color.
-    /// ## startingColorConstraints
-    /// The array of (Vertex#,Color) specifying the disallowed colors for vertices.
-    /// ## colorsRegister
-    /// The color register.
-    /// ## target
-    /// The target of the operation.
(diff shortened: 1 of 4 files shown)


samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -319,51 +149,30 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
+    /// ## nIterations
     /// An estimate of the maximum iterations needed.
     /// ## oracle
     /// The Oracle used to find solution.
+    /// ## statePrep
+    /// Routine that prepares an equal superposition of all basis states in the search space.
     ///
     /// # Output
     /// Int Array giving the color of each vertex.
     ///
```
cgranade, line 161:
```suggestion
```

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -319,51 +149,30 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
-        } until (correct or iter > maxIterations)
-        fixup {
-            set iter += 1;
-        }
-        if (not correct) {
-            fail "Failed to find a coloring.";
-        }
+        Message($"Trying search with {nIterations} iterations...");
+        ApplyGroversAlgorithmLoop(register, oracle, statePrep, nIterations);
+        let res = MultiM(register);
```
cgranade, line 173:
```suggestion
        let res = MeasureEachZ(register);
```

https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.measurement.measureeachz

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -319,51 +149,30 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
+        let coloring = MeasureColoring(bitsPerColor, register);
```
cgranade, line 174: This seems a bit odd, since you just measured in the immediately preceding line; is line 173 still needed, then? It looks like `res` is not used anywhere at the moment.

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +138,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+    }
+
     /// # Summary
     /// Estimate the number of iterations required for solution.
     ///
     /// # Input
-    /// ## nQubits
-    /// The number of qubits being used.
+    /// ## searchSpaceSize
+    /// The size of the search space
```
cgranade, line 229:
```suggestion
    /// The size of the search space.
```

samples/algorithms/sudoku-grover/SudokuQuantum.cs
```diff
@@ -131,6 +130,14 @@ struct EmptySquare
                                 emptySquareEdges.Add(
                                     (emptyIndex, emptyIndexes[i, jj]));
                         }
+                        if (size == 9)
+                        { // Invalidate numbers that are illegal on a 9x9 board
+                            for (int invalid = 9; invalid < 16; invalid++)
```
cgranade, line 135:
```suggestion
                            foreach (var invalid in Enumerable.Range(9, 16 - 9))
```

cgranade (commented):
Did an initial pass, and these look like some nice improvements; thanks for opening the PR! There's a few things that may be good to address, then I'm happy to do a more detailed pass. Thanks again!

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
adrianleh: These functions look identical in the documentation - is the one suggested more idiomatic?

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -111,27 +113,20 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
-            // startNumberConstraints. However, this did not scale well with 
-            // the simulator, and so instead we use 
-            // ApplyVertexColoringOracle4Bit9Color which has the 9 color 
-            // restriction built in.
-            set oracle = ApplyVertexColoringOracle4Bit9Color(numVertices, emptySquareEdges, startingNumberConstraints, _, _);
-        } elif (size != 4) {
+        let oracle = ApplyVertexColoringOracle(numVertices, bitsPerColor, startingNumberConstraints, _, _);
+        let statePrep = PrepareSearchStatesSuperposition(numVertices, bitsPerColor, startingNumberConstraints, _);
+        let searchSpaceSize = SearchSpaceSize(numVertices, bitsPerColor, startingNumberConstraints);
+        if (size != 4 and size != 9) {
```
tcNickolas, line 119: I'd move this check to be the first thing in the operation. Nothing bad will happen from defining operations for other sizes, but logically we want to do this check before we proceed with the solution

tcNickolas (commented):
I would also update the sample puzzles used by Program.cs. 
* Since the sparse simulator does so much better than the full state one, we can switch `4x4-classic` to use quantum solution as well, and add another 4x4 example in which even more digits are missing (I think it can do 12 missing digits, right?)
* I'm not sure `9x9-64` can be solved using quantum even with sparse simulator, I would drop it altogether and update `9x9-2` (or add `9x9-3`) to include the largest case we can solve quantum with sparse simulator

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -113,257 +113,66 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
         colorsRegister : Qubit[],
         target : Qubit
     )
     : Unit is Adj + Ctl {
         let nEdges = Length(edges);
-        let nStartingColorConstraints = Length(startingColorConstraints);
-        // we are looking for a solution that:
-        // (a) has no edge with same color at both ends and
-        // (b) has no Vertex with a color that violates the starting color constraints.
+        // we are looking for a solution that has no Vertex with a color that violates the starting color constraints.
```
tcNickolas, line 125: I think for this oracle we need to keep the other comment - we encode the prohibited colors in the search space, so the oracle only looks for a solution that has no edge connecting two vertices of the same color

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -113,257 +113,66 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
-    ///                    (1, 3),(2, 1),(2, 3),(2, 0),(3, 2),(3, 0),(3, 1)]
-    /// This is saying that vertex 0 is not allowed to have values 1,3,2
-    /// and vertex 1 is not allowed to have values 2,0,3
-    /// and vertex 2 is not allowed to have values 1,3,0
-    /// and vertex 3 is not allowed to have values 2,0,1
-    /// A valid graph coloring solution is: [0,1,2,3]
-    /// i.e. vertex 0 has color 0, vertex 1 has color 1 etc.
     operation ApplyVertexColoringOracle (
-        numVertices : Int, bitsPerColor : Int, edges : (Int, Int)[],
-        startingColorConstraints : (Int, Int)[],
```
tcNickolas, line 135: Since we dropped this parameter, we need to remove the `## startingColorConstraints` entry in documentation comments as well (GitHub doesn't allow me to leave this comment there)

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -113,257 +113,66 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
     ///
     /// # Input
-    /// ## numVertices
+    /// ## nVertices
     /// The number of Vertices in the graph.
     /// ## bitsPerColor
     /// The number of bits per color.
-    /// ## maxIterations
+    /// ## nIterations
     /// An estimate of the maximum iterations needed.
```
tcNickolas, line 153: Since we're updating the meaning of the parameter from the maximum number of iterations needed to the exact number done, let's also update the documentation

samples/algorithms/sudoku-grover/Program.cs
```diff
@@ -151,19 +164,19 @@ static void Main(string[] args)
...
+                    { 8,4,5, 0,0,0, 0,0,0 },
+                    { 7,0,8, 2,0,1, 3,5,0 },
+                    { 5,2,6, 4,0,3, 8,9,1 },
+                    { 1,3,4, 5,0,0, 0,0,0 },
+                    { 4,6,9, 1,2,8, 7,3,5 },
+                    { 2,8,7, 3,5,6, 1,4,9 },
+                    { 3,5,1, 9,4,7, 6,0,8} };
                 int[,] puzzle9_copy = CopyIntArray(puzzle9);
 
                 Console.WriteLine("Solving 9x9 with 64 missing numbers using classical computing.");
```
tcNickolas, line 182: Let's update the log message to match the update in the puzzle (both here and on line 187)

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +137,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+    /// # Inputs
+    /// ## nVertices
+    /// The number of vertices in the graph.
+    /// ## bitsPerColor
+    /// The bit width for number of colors.
+    /// ## startingNumberConstraints
+    /// The array of (Vertex#, Color) specifying the disallowed colors for vertices.
+    ///
+    /// # Output
+    /// A 2D array of amplitudes where the first index is the cell and the second index is the value of a basis state (i.e., value) for the cell
```
tcNickolas, line 153: It also makes sense to mention what are the values of the array elements - 1.0 for allowed values and 0.0 for prohibited ones

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +137,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+    function AllowedAmplitudes(
+        nVertices : Int,
+        bitsPerColor : Int,
+        startingNumberConstraints : (Int, Int)[]
+    ) : Double[][] {
+        mutable amplitudes = [[1.0, size=1 <<< bitsPerColor], size=nVertices];
+        for (cell, value) in startingNumberConstraints {
```
tcNickolas, line 160: You should be able to ignore the second element of the tuple, since the assumption is that all prohibited colors provided to the solver are unique and within the valid range of colors
```suggestion
        for (cell, _) in startingNumberConstraints {
```

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +137,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+        // For each vertex, create an array of possible states we're looking at.
+        let amplitudes = AllowedAmplitudes(nVertices, bitsPerColor, startingNumberConstraints);
+        // For each vertex, prepare a superposition of its possible states on the chunk storing its color.
+        for (amps, chunk) in Zipped(amplitudes, colorRegisters) {
+            PrepareArbitraryStateD(amps, LittleEndian(chunk));
+        }
+    }
+
+    /// # Summary
+    /// Prepare equal superposition of all basis states that satisfy the constraints
```
tcNickolas, line 198: This summary looks like it is from the `PrepareSearchStatesSuperposition` operation rather than this function

tcNickolas (commented):
I made another pass, once the comments are addressed, should be good!

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -30,8 +30,8 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
     /// ## register
     /// The register of qubits to be measured.
     operation MeasureColoring (bitsPerColor : Int, register : Qubit[]) : Int[] {
-        let numVertices = Length(register) / bitsPerColor;
-        let colorPartitions = Partitioned(ConstantArray(numVertices - 1, bitsPerColor), register);
+        let nVertices = Length(register) / bitsPerColor;
+        let colorPartitions = Partitioned(ConstantArray(nVertices - 1, bitsPerColor), register);
```
cgranade, line 34: Minor, but since this was written, the new `[foo, size=n]` syntax was introduced. While updating, it may be helpful to use that notation here in place of `ConstantArray`:

```suggestion
        let colorPartitions = Partitioned([bitsPerColor, size=nVertices - 1], register);
```

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -60,23 +60,20 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
     /// ## edges
     /// The array of (Vertex#,Vertex#) specifying the Vertices that can not be
     /// the same color.
-    /// ## startingColorConstraints
-    /// The array of (Vertex#,Color) specifying the disallowed colors for vertices.
     ///
     /// # Output
-    /// A unitary operation that applies `oracle` on the target register if the control
-    /// register state corresponds to the bit mask `bits`.
+    /// An marking oracle that with the purpose of marking states as allowed where the colors of qubits related by an edge constraints are not equal.
```
cgranade, line 75: I'm not sure this is gramatically correct? "That with" feels awkward at the least.

```suggestion
    /// An marking oracle that marks as allowed those states in which the colors of qubits related by an edge constraint are not equal.
```

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -113,259 +110,67 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
         colorsRegister : Qubit[],
         target : Qubit
     )
     : Unit is Adj + Ctl {
         let nEdges = Length(edges);
-        let nStartingColorConstraints = Length(startingColorConstraints);
-        // we are looking for a solution that:
-        // (a) has no edge with same color at both ends and
-        // (b) has no Vertex with a color that violates the starting color constraints.
+        // we are looking for a solution that has no edge with same color at both ends and
```
cgranade, line 122: This comment looks to end suddenly?

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -104,34 +106,26 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
-            // Although we could use ApplyVertexColoringOracle for 9x9, we would
-            // have to add restrictions on each color to not allow colors 8 
-            // through 15. This could be achieved by adding these to 
-            // startNumberConstraints. However, this did not scale well with 
-            // the simulator, and so instead we use 
-            // ApplyVertexColoringOracle4Bit9Color which has the 9 color 
-            // restriction built in.
-            set oracle = ApplyVertexColoringOracle4Bit9Color(numVertices, emptySquareEdges, startingNumberConstraints, _, _);
-        } elif (size != 4) {
+        if (size != 4 and size != 9) {
```
cgranade, line 114: You may not need parentheses here; removing them may suppress a warning.

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -104,34 +106,26 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
             fail $"Cannot set size {size}: only a grid size of 4x4 or 9x9 is supported";
         }
-        let numIterations = NIterations(bitsPerColor * numVertices);
-        Message($"Running Quantum test with #Vertex = {numVertices}");
+        let bitsPerColor = size == 9 ? 4 | 2;
+        let oracle = ApplyVertexColoringOracle(nVertices, bitsPerColor, emptySquareEdges, _, _);
+        let statePrep = PrepareSearchStatesSuperposition(nVertices, bitsPerColor, startingNumberConstraints, _);
+        let searchSpaceSize = SearchSpaceSize(nVertices, bitsPerColor, startingNumberConstraints);
+        let numIterations = NIterations(searchSpaceSize);
+        Message($"Running Quantum test with #Vertex = {nVertices}");
```
cgranade, line 122:
```suggestion
        Message($"Running Quantum test with # of vertices = {nVertices}");
```

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +137,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+    /// # Summary
+    /// Encodes stating number constraints into amplitudes.
+    ///
+    /// # Inputs
+    /// ## nVertices
+    /// The number of vertices in the graph.
+    /// ## bitsPerColor
+    /// The bit width for number of colors.
+    /// ## startingNumberConstraints
+    /// The array of (Vertex#, Color) specifying the disallowed colors for vertices.
```
cgranade, line 150: Adding an `# Example` or `# Remarks` section may be helpful here.

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +137,102 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+        startingNumberConstraints : (Int, Int)[],
+        register : Qubit[]
+    ) : Unit is Adj + Ctl {
+        // Split the given register into nVertices chunks of size bitsPerColor.
+        let colorRegisters = Chunks(bitsPerColor, register);
+        // For each vertex, create an array of possible states we're looking at.
+        let amplitudes = AllowedAmplitudes(nVertices, bitsPerColor, startingNumberConstraints);
+        // For each vertex, prepare a superposition of its possible states on the chunk storing its color.
+        for (amps, chunk) in Zipped(amplitudes, colorRegisters) {
+            PrepareArbitraryStateD(amps, LittleEndian(chunk));
```
cgranade, line 194: This could be a relatively expensive operation, depending. If gate depth is getting to be unwieldy, it may be good to investigate replacing this by something that prepares the state more directly, using your prior knowledge about the structure of that state.

cgranade (commented):
Looking a lot better, thanks for your edits! I think the main thing left that I might raise would be the state preparation using arbitrary prep; that definitely works, but may be worth a double check to ensure that it results in practical depths.

It also appears like some changes may be missing from the PR?

Anyway, this is really coming along, thanks for your improvements to this sample!

adrianleh:
@cgranade Could you elaborate which parts are missing?

samples/algorithms/sudoku-grover/Sudoku.qs
adrianleh: I don't think this is a big issue, as on simulators adding a bit of depth doesn't have a major impact and on hardware we are only able to run very limited instances where the state prep has not caused issues in practice

samples/algorithms/sudoku-grover/ColoringGroverWithConstraints.qs
```diff
@@ -113,257 +113,66 @@ namespace Microsoft.Quantum.Samples.ColoringGroverWithConstraints {
...
-                within {
-                    ApplyOrOracle(color[0 .. 2], tempQubit);
-                } apply{
-                    // AND color's most significant bit with OR of least significant bits.
-                    // This will set conflictQubit to 1 if color > 8.
-                    CCNOT(color[3], tempQubit, conflictQubit);
-                }
+            for ((start, end), conflictQubit) in Zipped(edges, edgeConflictQubits) {
+                // Check that endpoints of the edge have different colors:
+                // apply ColorEqualityOracle_Nbit oracle;
```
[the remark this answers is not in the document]
cgranade: @adrianleh: It looks like this still uses the old name? Was that intentional?

cgranade (commented):
Left a few more comments, but it looks like this is getting close, thank you! 

Overall, I would suggest that if you disagree with a comment, leaving another comment to that effect can help others understand why a conversation was resolved.

samples/algorithms/sudoku-grover/Sudoku.qs
adrianleh: `value` is referenced in the loop body hence I didn't apply the suggestion

samples/algorithms/sudoku-grover/Sudoku.qs
```diff
@@ -143,21 +138,109 @@ namespace Microsoft.Quantum.Samples.SudokuGrover {
...
+    /// # Inputs
+    /// ## nVertices
+    /// The number of vertices in the graph.
+    /// ## bitsPerColor
+    /// The bit width for number of colors.
+    /// ## startingNumberConstraints
+    /// The array of (Vertex#, Color) specifying the disallowed colors for vertices.
+    ///
+    /// # Remarks
+    /// Creates the search space. Using the allowed amplitudes prepares uniform superposition of all allowed values for each cell
```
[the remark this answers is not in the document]
adrianleh: You're absolutely right. Will change that

cgranade:
It looks like the Markdown link check failed due to a transient network error, so should be OK to proceed there.

adrianleh:
Thank you for looking at this, I have just added another commit to fix an issue that exists in the proposed and existing code where edge constraints are not created for subgrid relations

cgranade (approved):

PR #697, opened 2022-07-13
Update quantinuum hello world samples

achocron:


samples/azure-quantum/hello-world/HW-quantinuum-cirq.ipynb
```diff
@@ -168,11 +169,12 @@
     }
    },
    "source": [
-    "The circuit you built is a simple quantum random bit generator. With Quantinuum's API validator, we will be able to confirm that the circuit is able to be run on their H1 emulator and hardware."
+    "The circuit you built is a simple quantum random bit generator. With Quantinuum's Syntax Checker, we will be able to confirm that the circuit is able to be run on their H1 emulator and hardware."
```
cgranade, line 172: Especially given the new name, we should emphasize that `apival` is not a simulator and in particular always returns `Zero` for all measurements. I think we do that below, but stating it earlier as well may help avoid confusion.

samples/azure-quantum/hello-world/HW-quantinuum-qiskit.ipynb
```diff
@@ -152,36 +203,47 @@
     }
    },
    "source": [
-    "The circuit you built is a simple quantum random bit generator. With Quantinuum's API validator, we will be able to confirm that the circuit is able to be run on their H1 emulator and hardware."
+    "The circuit you built is a simple quantum random bit generator. With Quantinuum's Syntax Checker, we will be able to confirm that the circuit is able to be run on the Quantinuum H1-2 emulator and hardware."
```
cgranade, line 206: Similar comment here.

cgranade (approved):

PR #700, opened 2022-07-21
Enable verbose logging from test.ps1.

cgranade:
This PR makes more verbose logging available from test.ps1.


ricardo-espinoza (approved):

cgranade:
The Markdown link check is due to microsoft.com being down with a timeout; I think it's fair to dismiss that broken link detection as a false positive and to keep going.

PR #704, opened 2022-07-29
Update Deutsch–Jozsa with Noise.ipynb

achocron:
The workflow file changes in this PR are due to an issue with samples that have spaces in the names or special characters (like the long dash).

### DIFF
--- a/.github/workflows/hosted-notebook-samples-check.yml
+++ b/.github/workflows/hosted-notebook-samples-check.yml
@@ -33,7 +33,7 @@ jobs:
             # Restore spaces so they don't get double encoded
             file=$(echo $file | sed 's/%20/ /g')
             echo "Getting download link for $file"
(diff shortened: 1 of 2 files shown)


.github/workflows/hosted-notebook-samples-check.yml
```diff
@@ -17,14 +17,18 @@ jobs:
         id: prepare
         run: |
           FILES=$(git diff origin/${{ github.event.pull_request.base.ref }} --name-only)
+          FILES=$(echo $FILES | awk -F\" '{OFS="\"";for(i=2;i<NF;i+=2)gsub(/ /,"%20",$i);print}')
```
achocron, line 20: This change is so that files with spaces in the names are also checked to be tested

cgranade (approved):

PR #726, opened 2022-10-08
Updating samples to September 2022 Hot Fix 1 release build 0.26.233415

ricardo-espinoza:


anjbur (approved):
Comparing this to the previous PR #724, it looks like this is missing an update in [this file](https://github.com/microsoft/Quantum/blob/main/samples/qir/oracle-generator/qsharp/project.csproj). Otherwise, looks good to me!

PR #733, opened 2022-10-31
Resource estimation samples

msoeken:
`estimation-qsharp.ipynb`

**Title:** Estimates with Q# input

**Brief description:** Estimate and analyze the physical resource estimates of a quantum program written in Q# and targeting fault-tolerant quantum computers.

---

`advanced-estimation.ipynb`

**Title:** Advanced analysis of estimates

**Brief description:** Take results from multiple resource estimates, visualize, and compare their results.

---

`estimation-qiskit.ipynb`

**Title:** Estimates with Qiskit input

**Brief description:** Estimate and analyze the physical resource estimates of a quantum program written in Qiskit and targeting fault-tolerant quantum computers.

---

`estimation-qir.ipynb`

**Title:** Estimates with tools producing QIR

**Brief description:** Generate and submit QIR to the Resource Estimator.


samples/azure-quantum/resource-estimation/re-with-qir.ipynb
```diff
@@ -0,0 +1,168 @@
...
+    "job = resource_estimation_job_from_qir(provider, module.bitcode(), errorBudget=0.01)\n",
+    "job_monitor(job)\n",
+    "result = job.result()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Finally, we print the resource estimation table."
```
viktorveis, line 148: QIR is hidden from users in this example. I suggest that we have a cell that prints out generated QIR.

msoeken: The ASCII version of the bitcode is now printed in an additional cell before the job is submitted. There is also a comment that shows how to generate QIR bitcode from any QIR, which might not be generated from PyQIR.

samples/azure-quantum/resource-estimation/re-with-qir.ipynb
```diff
@@ -0,0 +1,194 @@
+{
```
xinyi-joffre, line 1: Once names of the samples are decided, we may want to align the file names themselves, so from maintenance perspective, it is also easier to know which files map to which samples in this repo (other than looking at mapping file).

I think this one is currently called something similar to re-advanced-analysis,ipynb

samples/azure-quantum/resource-estimation/hello-re-qiskit.ipynb
```diff
@@ -0,0 +1,600 @@
+{
```
xinyi-joffre, line 1: Similar comment as elsewhere in PR, can we name these similarly to their final name in gallery? So if not using hello world in sample title, we may want to rename file to just be re-qiskit.ipynb

samples/azure-quantum/resource-estimation/re-with-qir.ipynb
msoeken: Makes sense. But I think this comment was meant to be added to what is called now `re-experiments-qsharp.ipynb`.

xinyi-joffre (approved):

PR #761, opened 2023-01-19
Adding storage sample for gallery

anraman:
This sample shows you how to connect and use Azure Quantum with external datasources such as [Azure Blob Storage](https://learn.microsoft.com/azure/storage/blobs/storage-blobs-introduction). In particular, how to connect to the attached blob storage account associated with each AQ Workspace. Also shows how to load files from a URL and work with local storage.

### DIFF
--- a/binder-index.md
+++ b/binder-index.md
@@ -252,6 +252,14 @@ These are noted in the README.md files for each sample, along with complete inst
       <td></td>
     </tr>
     <tr>
+      <td></td>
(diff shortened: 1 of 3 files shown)


anraman:
^check fails because I added a pre-emptive link to the final location of the Jupyter nb file which obviously won't exist until this PR is merged: https://github.com/microsoft/quantum/blob/main/samples/azure-quantum/utilities/storage-data-management.ipynb

xinyi-joffre (approved):

PR #766, opened 2023-01-25
update ml wine sample miss rate to match halfmoon

wadeschulz:
Updated the .NET host script and the Q# Training script in the ML wine sample to be consistent with both the Half Moon examples and the python host by reporting percent miss rate instead of number of missed classifications


wadeschulz:
@microsoft-github-policy-service agree [company="Wade Schulz"]

wadeschulz:
@microsoft-github-policy-service agree

PR #788, opened 2023-03-07
Fixed the Docker image.

### LINKED ISSUE #787: "Prebuild Docker image" Action is broken.
[Actions/"Prebuild Docker image"](https://github.com/microsoft/Quantum/actions/workflows/prebuild-docker.yml).

kuzminrobin:
Fixes #787

### DIFF
--- a/.github/workflows/prebuild-docker.yml
+++ b/.github/workflows/prebuild-docker.yml
@@ -19,6 +19,14 @@ jobs:
     - name: Docker Login
       run: az acr login -n qdkimages
     - name: Build and push Docker image
+      # Would be nice to invoke "test-docker-image-build.yml" workflow here to avoid the code duplication,
+      # but the other workflows can only be invoked in a separate job (with no extra steps),
+      # rather than in a separate step.
+      # And different jobs use different runners (build agents), i.e. if one job invokes a workflow
+      # then the subsequent job cannot just use the binaries produced by that workflow.
+      # One job can only pass the binaries to the other job by means of artifacts, upload by one job and
+      # download by another job. But artifacts upload is alien for that workflow.
+      # https://docs.github.com/actions/using-workflows/storing-workflow-data-as-artifacts
       run: |
         $Now = [DateTime]::Now;
         $ImageTag = "${{ github.sha }}"
--- a/.github/workflows/test-docker-image-build.yml
+++ b/.github/workflows/test-docker-image-build.yml
@@ -1,0 +1,23 @@
+# https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions
+
+name: Test Docker image build
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build Docker image
+      run: |
+        $ImageTag = "${{ github.sha }}"
+        $RepoName = "public/quantum/samples"
+
+        $LocalTag = "${RepoName}:${ImageTag}"
+        $RemoteRepo = "${{ secrets.ACR_REGISTRY }}/${RepoName}"
+
+        docker build Build/images/samples --tag $LocalTag
+      shell: pwsh
--- a/.github/workflows/test-docker-image-build.yml
+++ b/.github/workflows/test-docker-image-build.yml
@@ -17,7 +17,6 @@ jobs:
         $RepoName = "public/quantum/samples"
 
         $LocalTag = "${RepoName}:${ImageTag}"
-        $RemoteRepo = "${{ secrets.ACR_REGISTRY }}/${RepoName}"
 
         docker build Build/images/samples --tag $LocalTag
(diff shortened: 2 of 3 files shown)


.github/workflows/prebuild-docker.yml
```diff
@@ -2,9 +2,10 @@ name: Prebuild Docker image
 
 on:
   push:
-    branches:
-    - main
-
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
billti, line 7: Do we want to rebuild the docker image on all pull requests too?

kuzminrobin: It is an 8-minunte action that detects the breaking changes. I thought, why not.
Previously the break stayed unnoticed for 3 months.

Build/images/samples/Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN dotnet tool install --global PowerShell --version 7.2.4
 # Binder.
 RUN pip install jupytext
 # Make the C# and PowerShell kernels available to Jupytext as well.
-RUN dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive && \
+RUN dotnet tool install -g --add-source "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" Microsoft.dotnet-interactive --version 1.0.355307 && \
```
ScottCarda-MS, line 51: Do we know why it doesn't just grab the latest version by default? Do we have to specify it here?

kuzminrobin: The command does grab the latest version by default, but the versions after this one have dropped the support of .NET 6.0. 
See [Versions](https://www.nuget.org/packages/Microsoft.dotnet-interactive/1.0.355307#versions-body-tab) / {Click a specific version} / {see the [Frameworks](https://www.nuget.org/packages/Microsoft.dotnet-interactive/1.0.355307#supportedframeworks-body-tab) tab for that version}.

I can add this explanation as the code comments. Shall I add?

ScottCarda-MS: Yah having a comment explaining the reason for the specific version would be good so that devs in the future looking at this know why that version specifically and don't updated it to the latest.

kuzminrobin: Done.

.github/workflows/prebuild-docker.yml
kuzminrobin: Added a new GitHub action triggered in the PR.

.github/workflows/test-docker-image-build.yml
```diff
@@ -0,0 +1,23 @@
+# https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions
+
+name: Test Docker image build
```
ScottCarda-MS, line 3: Is the idea here to just check to make sure the docker image can be built? This just seems to build the image and then forget about it.

.github/workflows/test-docker-image-build.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build Docker image
+      run: |
+        $ImageTag = "${{ github.sha }}"
+        $RepoName = "public/quantum/samples"
+
+        $LocalTag = "${RepoName}:${ImageTag}"
+        $RemoteRepo = "${{ secrets.ACR_REGISTRY }}/${RepoName}"
```
ScottCarda-MS, line 20: Where does this variable get used? Does it get used in a different context after this job?

.github/workflows/test-docker-image-build.yml
billti: I think this is the intent. Does it block the PR or merging if it fails?

kuzminrobin: Confirmed. The idea here is to make sure the docker image can be built, and _**if not then to block the PR**_.

I didn't actually check if the failure really blocks the PR, I will do it in my [side draft PR](https://github.com/microsoft/Quantum/pull/789). Good point.

.github/workflows/test-docker-image-build.yml
kuzminrobin: It is a copy-paste problem, good find.

It was used in the fragment where I copied this from. I will remove this var.

.github/workflows/test-docker-image-build.yml
kuzminrobin: Confirmed, the [failure does block](https://github.com/microsoft/Quantum/pull/789) the PR.

.github/workflows/test-docker-image-build.yml
kuzminrobin: Removed.

kuzminrobin:
PLease review again.

ScottCarda-MS (approved):

PR #791, opened 2023-03-09
Removing local resources estimator

DmitryVasilevsky:
Removed use of deprecated ResourcesEstimator class.

### DIFF
--- a/samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
+++ b/samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
@@ -12,10 +12,10 @@ using Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators;
 
 namespace Microsoft.Quantum.Simulation.Simulators
 {
-    public class FlameGraphResourcesEstimator : ResourcesEstimator
+    public class FlameGraphResourcesEstimator : QCTraceSimulator
     {
         private FlameGraphCounter operationsCounter;
-        public static new QCTraceSimulatorConfiguration RecommendedConfig() =>
+        public static QCTraceSimulatorConfiguration RecommendedConfig() =>
             new QCTraceSimulatorConfiguration
             {
                 CallStackDepthLimit = Int32.MaxValue,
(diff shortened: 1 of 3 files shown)


samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
```diff
@@ -12,10 +12,10 @@
 
 namespace Microsoft.Quantum.Simulation.Simulators
 {
-    public class FlameGraphResourcesEstimator : ResourcesEstimator
+    public class FlameGraphResourcesEstimator : QCTraceSimulator
```
swernli, line 15: While it makes sense that this resolves the compilation error, is this actually valid given what the sample is? Does `FlameGraphResourceEstimator` produce valid/usable results when run through the tracer? If not, maybe the whole sample should just be removed?

samples/algorithms/integer-factorization/QuantumVizEstimator.cs
```diff
@@ -12,10 +12,10 @@
 
 namespace Microsoft.Quantum.Simulation.Simulators
 {
-    public class QuantumVizEstimator : ResourcesEstimator
+    public class QuantumVizEstimator : QCTraceSimulator
```
swernli, line 15: Same thing here, perhaps this would be better removed if the tracer behavior is too different?

swernli (commented):
This seems good enough to compile, but I'm not sure if it changes the behavior of the samples too much. Perhaps we should consider removing these two samples entirely.

DmitryVasilevsky:
> This seems good enough to compile, but I'm not sure if it changes the behavior of the samples too much. Perhaps we should consider removing these two samples entirely.

It still produces the results, but results are less aggregated and therefore less usable. Since we are keeping QCTraceSimulator I think it makes sense to keep these samples.

samples/algorithms/integer-factorization/FlameGraphResourcesEstimator.cs
DmitryVasilevsky: In this particular instance no additional functionality of ResourcesEstimator is used. QCTraceSimulator works just as well. But see other notes.

samples/algorithms/integer-factorization/QuantumVizEstimator.cs
DmitryVasilevsky: In this particular instance no additional functionality of ResourcesEstimator is used. QCTraceSimulator works just as well. But see other notes.

swernli (approved):

PR #801, opened 2023-04-12
Add Sessions introduction sample

ashwinmayya:
Add Sessions introduction sample


masenol (approved):
Just did a quick skim through the sample, looks good! My assumption is that we'll get a sign off from PM side prior to releasing this. Approving.

l-enloe (approved):

PR #803, opened 2023-04-24
Remove Aspen M-2 references from Rigetti samples

masenol:


achocron (approved):
Looks good; let's also make sure this is tested in the portal

PR #813, opened 2023-07-08
Add 64-bit Int input support and arithmetics op in read_qir.hpp

itungc:
Add 64-bit integer input support and arithmetic function (SRem, addition, multiplication) support to read_qir.hpp in the oracle-generator.


itungc:
@microsoft-github-policy-service agree

PR #829, opened 2023-09-15
Add a generalised notebook to approximate Energy using Jordan-Wigner transformation for any molecule

rheaparekh:
This PR is the result of the hackathon project: https://hackbox.microsoft.com/hackathons/hackathon2023/project/28045

Here we change the 'Approximate Energy using Jordan-Wigner transformation" notebook to be generalised for any molecule. This notebook was made to calculate the ground state energy for H2, however we have as a part of the hackathon project, generalised this notebook where it would take in the input of any molecule as a yaml file from the NWChem library: https://learn.microsoft.com/en-us/azure/quantum/user-guide/libraries/chemistry/samples/end-to-end, and provide the results for the ground state energy for that molecule.


jkingdon-ms:
@ffrachon if you have a moment to review please

PR #831, opened 2023-09-19
Add RUS Azure Quantum notebook sample

swernli:
This adds a notebook with a subset of the RUS experiment on Quanitnuum to allow users to repeat this experiment on their own as a hosted notebook.


ffrachon (approved):
Would be great to evolve this sample to take advantage of sessions (speeds up execution on Quantinuum), but not blocking.

